PR #22, opened 2026-06-07
feat: add qr contact sharing

clayrootkit:


xmreur (approved):
seems good, no issues and works both on linux and android (i'll assume windows is alright too)

PR #84, opened 2026-07-13
fix: replace CTE+self-join in getLastMessagePreviews with direct MAX …

clayrootkit:
…aggregate query to prevent spurious 'Deleted' previews

### DIFF
--- a/lib/database/messages.dart
+++ b/lib/database/messages.dart
@@ -993,26 +993,22 @@ class MessagesDb {
         );
       }
 
+      // SQLite guarantees that bare columns in a MAX-aggregate query come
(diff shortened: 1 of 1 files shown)


xmreur (approved):
Group Message Path (same function): Potential Issue
The group branch inside getLastMessagePreviews uses a Dart-side sort (orderBy: 'timestamp DESC') and takes the first unseen row per group . Unlike the direct query, this path does not use MAX() at the SQL level, it fetches all group messages sorted by timestamp and iterates in Dart. This is a pre-existing inefficiency (not introduced by this PR), but it means for large group histories it loads many rows into memory. Not a blocker, but worth a follow-up issue.

PR #85, opened 2026-07-13
fix: prevent double SafeArea stacking on mobile causing empty space i…

clayrootkit:
…n screens (Settings, Chat, Profile, etc.)


xmreur (commented):
needs testing before merge but looks good code-wise

PR #99, opened 2026-07-30
fix(group): resolve own sender identity from keystore

clayrootkit:
Own group messages showed "Cannot decrypt message" when reopening a chat. Group messages use the sender-key scheme, where the signature is verified against the sender's public key -- looked up in the `users` table. The local user's own row carries a placeholder identity (`identityJson: 'NONE'`), so signature verification failed for own messages only, while peers decrypted fine.

Resolve the signing keys from the keystore for the local user, since that is the key that actually produced the signature. Peer lookups are unchanged, and signature verification stays mandatory for every sender.

Sending was unaffected because outbound messages are appended in memory with the plaintext already available; the failure only surfaced once rows were re-read from the database.

### DIFF
--- a/lib/screens/group_chat.dart
+++ b/lib/screens/group_chat.dart
@@ -721,9 +721,10 @@ class _GroupChatScreenState extends State<GroupChatScreen> {
     required String wire,
     required String transportSenderId,
   }) async {
-    final senderKeys = await loadPeerIdentityFromDb(
+    final senderKeys = await loadGroupSenderIdentity(
       widget.keyManager,
       transportSenderId,
+      localUserId: widget.userId,
     );
     if (senderKeys == null) {
       throw ArgumentError('Unknown sender identity');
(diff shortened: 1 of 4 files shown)


xmreur (approved):
looks good

xmreur:
build:test

xmreur:
## PR build results

|---|---|---|
| Windows | failure | N/A |
| Linux | success | [Download](https://github.com/xmreur/prysm/actions/runs/30538906204/artifacts/8757926784) |
| macOS | success | [Download](https://github.com/xmreur/prysm/actions/runs/30538906204/artifacts/8758033544) |
| iOS | success | [Download](https://github.com/xmreur/prysm/actions/runs/30538906204/artifacts/8757992560) |
| Android | success | [Download](https://github.com/xmreur/prysm/actions/runs/30538906204/artifacts/8758082671) |

Triggered by: @xmreur
Run: https://github.com/xmreur/prysm/actions/runs/30538906204

PR #102, opened 2026-07-30
feat(chat): schedule messages for later delivery

clayrootkit:
Long-pressing the send button opens a date/time picker; the message is queued and delivered once its time arrives, in both direct chats and groups. A message whose time passed while the app was closed is sent on the next flush rather than dropped.

Queued bodies are encrypted for self at rest, so a pending message is never plaintext on disk. Pending messages are listed and cancellable from the per-chat settings screen.

Sends are driven by a dedicated one-shot timer in SyncCoordinator armed at the exact due instant, deliberately kept off flushAllPending: that method holds its _flushing guard for as long as the retry queues take, and a single unreachable peer can stall them for minutes.

Also stops HomeScreen.dispose() from shutting Tor down. The controller is owned by _MyAppState and outlives the widget, so a rebuild that remounts HomeScreen killed a Tor process the new instance still needed, which left every pending queue blocked. Real exit still stops Tor via quitApp() and the detached lifecycle branch.

### DIFF
--- a/lib/database/message_schema_migrations.dart
+++ b/lib/database/message_schema_migrations.dart
@@ -11,7 +11,7 @@ class MessageSchemaMigrations {
   MessageSchemaMigrations._();
 
   /// Current schema version. Bump alongside a new _upgradeToVN step.
-  static const int dbVersion = 11;
+  static const int dbVersion = 12;
 
   static Future<void> onCreate(Database db, int version) async {
     await _createV2(db);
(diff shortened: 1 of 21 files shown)


xmreur:
build:test

xmreur:
## PR build results

|---|---|---|
| Windows | failure | N/A |
| Linux | success | [Download](https://github.com/xmreur/prysm/actions/runs/30608998138/artifacts/8784760618) |
| macOS | success | [Download](https://github.com/xmreur/prysm/actions/runs/30608998138/artifacts/8784819317) |
| iOS | success | [Download](https://github.com/xmreur/prysm/actions/runs/30608998138/artifacts/8784854883) |
| Android | success | [Download](https://github.com/xmreur/prysm/actions/runs/30608998138/artifacts/8784910712) |

Triggered by: @xmreur
Run: https://github.com/xmreur/prysm/actions/runs/30608998138

xmreur (approved):

PR #128, opened 2026-08-06
fix(security): harden transport, db-at-rest, ratchet, group and privacy

MrBuondi:
Security hardening: inbound transport, databases at rest, ratchet, group channels and privacy.

## Changes
- **DB at rest**: SQLCipher encryption, per-install key in secure storage, plaintext migration
- **Transport**: WS peer-identity proof, validated accept header, rate limits, bounded request bodies, path-traversal rejection
- **Ratchet**: wire-scheme negotiation, legacy key-wrap support, one-time prekey reservation, session cleared on contact delete
- **Group**: sender-index mutex, stale/soft-deleted message filtering, replay and legacy-envelope rejection
- **Privacy**: profile fetches gated behind ingress/auth, online-status honoured, onion redaction in logs, backup/device-transfer disabled
- **Calls**: audio receive window advanced only after MAC verification; libopus fetched over Tor with pinned SHA-256
- **Ops**: per-install random Tor control password; containerized L3 end-to-end security harness

## Verification
`flutter analyze` clean; `flutter test` 898 passed, 1 skipped.

## Known behaviour, accepted

**A group invite from a peer who is not already a local contact is dropped.** Group control
messages are authenticated against the sender identity held in the local user store, and
`_resolveSenderIdentity` is DB-only by design: resolving an unknown sender over Tor would let
unauthenticated inbound traffic force an outbound `GET /profile` — the M2 oracle this PR closes.
The identity is required, not merely checked: `GroupCryptoV2.decryptControlPayload` cannot read
the payload without it.

The sender still receives a normal ack and the invitee sees nothing. That is the agreed
trade-off ([decision](https://github.com/xmreur/prysm/pull/128#issuecomment-5205552544)), pinned
by the `POLICY:` tests in `test/security/group_profile_fetch_gate_test.dart`. Surfacing it as a
pending invite is tracked in #129.

Also tracked separately: neither transport propagates the receiver's status, so a rejected
message is recorded as delivered (#130). Pre-existing, both transports, all message types.

## Scope note

"No pre-auth outbound fetch" holds for the **group** paths. The 1:1 path still resolves an
unknown sender's identity before verification (`direct_message_auth.dart` via
`PrysmServer.dart:73`); closing it changes new-contact bootstrap and is a separate decision.

## Summary by CodeRabbit

* **Security & Privacy**
  * Strengthened WebSocket and group-message authentication and replay protection.
  * Improved one-time-prekey handling when keys are unavailable or reservations fail.
  * Redacted onion addresses from logs and disabled Android backups and data transfers.
  * Prevented profile lookups for unauthenticated or rejected messages.

* **Reliability**
  * Improved message recovery, database migrations, session cleanup, and audio decryption handling.

* **Call Quality**
  * Added inbound call-offer rate limiting.

* **Testing**
  * Expanded automated coverage, including isolated Tor end-to-end security tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,3 +28,10 @@ jobs:
 
       - name: Test
         run: flutter test
+
+      - name: L3 harness coverage notice
+        run: |
+          echo "The real-Tor L3 E2E harness did NOT run in this job."
+          echo "It is gated behind PRYSM_E2E=1 (this job runs plain"
+          echo "'flutter test', so test/headless_inbound_e2e_test.dart"
+          echo "skips here). It must be run locally with .l3e2e/run.sh."
--- a/.l3e2e/.gitignore
+++ b/.l3e2e/.gitignore
@@ -1,0 +1,1 @@
+workdir/
--- a/.l3e2e/Containerfile
+++ b/.l3e2e/Containerfile
@@ -1,0 +1,45 @@
+# Prysm — L3 E2E isolated experiment environment
+#
+# Why a container: Prysm's TorManager._killOrphanTorForce runs `pkill -9 tor`
+# with no pattern as a fallback before every start (lib/util/tor_service.dart:488-505).
+# Inside a container the PID namespace confines that blast radius to the
+# container itself — the host's tor processes are unreachable.
+#
+# Base: ubuntu:24.04, glibc 2.39. Constraints:
+#   - bundled tor binary requires GLIBC_2.38 (objdump on tor_executable/tor)
+#   - libsqlite3.so (host-built) requires GLIBC_2.34
+#   - flutter_tester needs only libc/libm/libdl (fully static otherwise)
+FROM ubuntu:24.04
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Last group: runtime libraries for the bundled tor binary (ldd on
+# tor_executable/tor — it is NOT statically linked).
+RUN apt-get update && apt-get install -y --no-install-recommends \
+      bash ca-certificates curl git xz-utils unzip \
+      build-essential \
+      psmisc procps \
+      libevent-2.1-7t64 libssl3t64 liblzma5 libzstd1 zlib1g \
+      libcap2 libseccomp2 libsystemd0 libbrotli1 \
+    && rm -rf /var/lib/apt/lists/*
+
+# Official Flutter SDK, same version as the host toolchain and .dart_tool/version.
+ARG FLUTTER_VERSION=3.44.8
+RUN curl -fsSL -o /tmp/flutter.tar.xz \
+      "https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_${FLUTTER_VERSION}-stable.tar.xz" \
+    && tar -xJf /tmp/flutter.tar.xz -C /opt \
+    && rm /tmp/flutter.tar.xz
+
+ENV PATH="/opt/flutter/bin:${PATH}"
+RUN git config --global --add safe.directory /opt/flutter \
+    && flutter --disable-analytics \
+    && flutter precache --linux \
+    && chown -R ubuntu:ubuntu /opt/flutter
+
+# ubuntu:24.04 ships a default uid-1000 user 'ubuntu'; the host user is also
+# uid 1000, so the bind-mounted repo and pub cache stay readable/writable.
+USER ubuntu
+ENV HOME=/home/ubuntu \
+    PUB_CACHE=/home/ubuntu/.pub-cache
+
+WORKDIR /work
--- a/.l3e2e/Containerfile
+++ b/.l3e2e/Containerfile
@@ -38,11 +38,23 @@ RUN curl -fsSL -o /tmp/flutter.tar.xz \
 ENV PATH="/opt/flutter/bin:${PATH}"
 RUN git config --global --add safe.directory /opt/flutter \
     && flutter --disable-analytics \
-    && flutter precache --linux \
+    && flutter precache --linux
+
+# Bind-mount ownership: ubuntu:24.04 ships a default uid-1000 user
+# 'ubuntu', but the host user may be any uid. .l3e2e/run.sh passes the host
+# uid/gid as build args, and this step remaps 'ubuntu' to match, so the
+# bind-mounted /work and ~/.pub-cache stay readable/writable for the
+# container user. The step sits AFTER the Flutter install so changing the
+# args does not invalidate the ~1 GB tarball download layer; the chown of
+# /opt/flutter runs AFTER the remap or the SDK would keep the old uid.
+ARG HOST_UID=1000
+ARG HOST_GID=1000
+RUN if [ "${HOST_UID}" != "1000" ] || [ "${HOST_GID}" != "1000" ]; then \
+      groupmod -g "${HOST_GID}" ubuntu \
+      && usermod -u "${HOST_UID}" -g "${HOST_GID}" ubuntu; \
+    fi \
     && chown -R ubuntu:ubuntu /opt/flutter
 
-# ubuntu:24.04 ships a default uid-1000 user 'ubuntu'; the host user is also
-# uid 1000, so the bind-mounted repo and pub cache stay readable/writable.
 USER ubuntu
 ENV HOME=/home/ubuntu \
     PUB_CACHE=/home/ubuntu/.pub-cache
--- a/.l3e2e/Containerfile
+++ b/.l3e2e/Containerfile
@@ -24,9 +24,14 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && rm -rf /var/lib/apt/lists/*
 
 # Official Flutter SDK, same version as the host toolchain and .dart_tool/version.
+# SHA-256 from Google's releases_linux.json for this exact archive — verified
+# against storage.googleapis.com so a tampered or truncated download aborts the
+# build before extraction.
 ARG FLUTTER_VERSION=3.44.8
+ARG FLUTTER_SHA256=672089e001571a9fbb209a495c583580c0c6c73ef98999264ba07fa93ace332d
 RUN curl -fsSL -o /tmp/flutter.tar.xz \
       "https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_${FLUTTER_VERSION}-stable.tar.xz" \
+    && echo "${FLUTTER_SHA256}  /tmp/flutter.tar.xz" | sha256sum -c - \
     && tar -xJf /tmp/flutter.tar.xz -C /opt \
     && rm /tmp/flutter.tar.xz
 
--- a/.l3e2e/run.sh
+++ b/.l3e2e/run.sh
@@ -9,12 +9,27 @@ cd "$(dirname "$0")/.."
 IMAGE=prysm-l3e2e
 WORK=.l3e2e/workdir
 
-# Safety gate mirrored inside the container, but cheap to check here too:
-# the harness's victim TorManager will `pkill -9 tor` without a pattern if
-# its targeted kill finds nothing (lib/util/tor_service.dart:488-505).
+# Safety gate for running the harness OUTSIDE the container. Inside the
+# container the PID namespace IS the containment: TorManager's cleanup can
+# `pkill -9 tor` without a pattern (lib/util/tor_service.dart:497) and can
+# only see container PIDs. This check is not mirrored inside the container;
+# it protects the host case, where that pattern-less kill would hit a real
+# host tor.
 if pgrep -x tor >/dev/null 2>&1; then
-  echo "WARNING: a tor process is running on the HOST. The container cannot" >&2
-  echo "kill it (PID namespace), but abort if this surprises you." >&2
+  if [ "${PRYSM_L3E2E_ALLOW_HOST_TOR:-}" != "1" ]; then
+    echo "ERROR: a tor process is running on the HOST. The L3 harness's" >&2
+    echo "victim TorManager runs a pattern-less 'pkill -9 tor'" >&2
+    echo "(lib/util/tor_service.dart:497) whenever its targeted pkill finds" >&2
+    echo "nothing — the normal case — and inside the container the PID" >&2
+    echo "namespace contains it. On the host there is no containment, so" >&2
+    echo "this run would kill that host tor process. Run the harness in the" >&2
+    echo "container (this script's normal path), or set" >&2
+    echo "PRYSM_L3E2E_ALLOW_HOST_TOR=1 to continue with a warning." >&2
+    exit 1
+  fi
+  echo "WARNING: a tor process is running on the HOST; continuing because" >&2
+  echo "PRYSM_L3E2E_ALLOW_HOST_TOR=1 is set. The pattern-less 'pkill -9 tor'" >&2
+  echo "(lib/util/tor_service.dart:497) could kill it." >&2
 fi
 
 mkdir -p "$WORK"
--- a/.l3e2e/run.sh
+++ b/.l3e2e/run.sh
@@ -9,40 +9,41 @@ cd "$(dirname "$0")/.."
 IMAGE=prysm-l3e2e
 WORK=.l3e2e/workdir
 
-# Safety gate for running the harness OUTSIDE the container. Inside the
-# container the PID namespace IS the containment: TorManager's cleanup can
-# `pkill -9 tor` without a pattern (lib/util/tor_service.dart:497) and can
-# only see container PIDs. This check is not mirrored inside the container;
-# it protects the host case, where that pattern-less kill would hit a real
-# host tor.
-if pgrep -x tor >/dev/null 2>&1; then
-  if [ "${PRYSM_L3E2E_ALLOW_HOST_TOR:-}" != "1" ]; then
-    echo "ERROR: a tor process is running on the HOST. The L3 harness's" >&2
-    echo "victim TorManager runs a pattern-less 'pkill -9 tor'" >&2
-    echo "(lib/util/tor_service.dart:497) whenever its targeted pkill finds" >&2
-    echo "nothing — the normal case — and inside the container the PID" >&2
-    echo "namespace contains it. On the host there is no containment, so" >&2
-    echo "this run would kill that host tor process. Run the harness in the" >&2
-    echo "container (this script's normal path), or set" >&2
-    echo "PRYSM_L3E2E_ALLOW_HOST_TOR=1 to continue with a warning." >&2
-    exit 1
-  fi
-  echo "WARNING: a tor process is running on the HOST; continuing because" >&2
-  echo "PRYSM_L3E2E_ALLOW_HOST_TOR=1 is set. The pattern-less 'pkill -9 tor'" >&2
-  echo "(lib/util/tor_service.dart:497) could kill it." >&2
+# Host-tor safety lives in the harness, not here:
+# test/headless_inbound_e2e_test.dart refuses to start while a tor process
+# is running on the HOST (override: PRYSM_L3E2E_ALLOW_HOST_TOR=1). This
+# script's container path is always safe: the container's PID namespace IS
+# the containment — TorManager's cleanup can `pkill -9 tor` without a
+# pattern (lib/util/tor_service.dart:497) and can only see container PIDs,
+# and the image starts no tor of its own.
+
+# The image remaps its 'ubuntu' user to the host uid/gid (HOST_UID/HOST_GID
+# build args, below) so bind mounts stay writable. uid 0 cannot be served
+# that way — the image's root ids are reserved — so abort early instead of
+# building a subtly broken image.
+if [ "$(id -u)" -eq 0 ]; then
+  echo "ERROR: refusing to run as root (uid 0): the container remaps its" >&2
+  echo "'ubuntu' user to the host uid, and uid 0 is reserved for root —" >&2
+  echo "the resulting image would be broken. Run the L3 E2E harness as a" >&2
+  echo "non-root user." >&2
+  exit 1
 fi
 
 mkdir -p "$WORK"
 # Create the host-side pub cache before mounting it: if the docker daemon
-# created it, it would be root:root 755 and the uid-1000 container user
-# could not write to it (flutter pub get would fail).
+# created it, it would be root:root 755 and the container user (remapped
+# to the host uid via HOST_UID/HOST_GID) could not write to it (flutter
+# pub get would fail).
 mkdir -p "$HOME/.pub-cache"
 rsync -a --delete \
   --exclude '.dart_tool' \
   lib test assets pubspec.yaml pubspec.lock analysis_options.yaml \
   tor_executable packages "$WORK"/
 
-docker build -f .l3e2e/Containerfile -t "$IMAGE" .l3e2e
+docker build -f .l3e2e/Containerfile -t "$IMAGE" \
+  --build-arg HOST_UID="$(id -u)" \
+  --build-arg HOST_GID="$(id -g)" \
+  .l3e2e
 
 # --network bridge: outbound internet only (Tor bootstrap needs it).
 # No host ports published: victim/attacker talk over the real Tor network.
--- a/.l3e2e/run.sh
+++ b/.l3e2e/run.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+# Prysm — L3 E2E runner: stages a repo copy and runs the harness in an
+# isolated container. The live repo is never touched by the container:
+# .dart_tool, package_config.json and native builds are regenerated inside
+# the staging dir (.l3e2e/workdir), which persists between runs for caching.
+set -euo pipefail
+cd "$(dirname "$0")/.."
+
+IMAGE=prysm-l3e2e
+WORK=.l3e2e/workdir
+
+# Safety gate mirrored inside the container, but cheap to check here too:
+# the harness's victim TorManager will `pkill -9 tor` without a pattern if
+# its targeted kill finds nothing (lib/util/tor_service.dart:488-505).
+if pgrep -x tor >/dev/null 2>&1; then
+  echo "WARNING: a tor process is running on the HOST. The container cannot" >&2
+  echo "kill it (PID namespace), but abort if this surprises you." >&2
+fi
+
+mkdir -p "$WORK"
+rsync -a --delete \
+  --exclude '.dart_tool' \
+  lib test assets pubspec.yaml pubspec.lock analysis_options.yaml \
+  tor_executable packages "$WORK"/
+
+docker build -f .l3e2e/Containerfile -t "$IMAGE" .l3e2e
+
+# --network bridge: outbound internet only (Tor bootstrap needs it).
+# No host ports published: victim/attacker talk over the real Tor network.
+exec docker run --rm \
+  -e PRYSM_E2E=1 \
+  -v "$PWD/$WORK":/work \
+  -v "$HOME/.pub-cache":"$HOME/.pub-cache":rw \
+  -e PUB_CACHE="$HOME/.pub-cache" \
+  "$IMAGE" bash -lc '
+    set -e
+    flutter pub get
+    flutter test test/headless_inbound_e2e_test.dart --timeout=45m --reporter expanded
+  '
--- a/.l3e2e/run.sh
+++ b/.l3e2e/run.sh
@@ -33,6 +33,10 @@ if pgrep -x tor >/dev/null 2>&1; then
 fi
 
 mkdir -p "$WORK"
+# Create the host-side pub cache before mounting it: if the docker daemon
+# created it, it would be root:root 755 and the uid-1000 container user
+# could not write to it (flutter pub get would fail).
+mkdir -p "$HOME/.pub-cache"
 rsync -a --delete \
   --exclude '.dart_tool' \
   lib test assets pubspec.yaml pubspec.lock analysis_options.yaml \
@@ -45,8 +49,7 @@ docker build -f .l3e2e/Containerfile -t "$IMAGE" .l3e2e
 exec docker run --rm \
   -e PRYSM_E2E=1 \
   -v "$PWD/$WORK":/work \
-  -v "$HOME/.pub-cache":"$HOME/.pub-cache":rw \
-  -e PUB_CACHE="$HOME/.pub-cache" \
+  -v "$HOME/.pub-cache":/home/ubuntu/.pub-cache:rw \
   "$IMAGE" bash -lc '
     set -e
     flutter pub get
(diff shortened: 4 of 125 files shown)


xmreur:
The PR contains several valuable security fixes, but I found issues that should be addressed before approval:

1. **Group replay protection is a high-water mark, not a replay window.** `recordInboundIndex` rejects every index lower than `maxSeen + 1`, so legitimate out-of-order messages can be dropped (for example, index 10 arriving before index 9). Either implement a bounded sliding window with replay tracking, or document and prove that strict ordering is guaranteed by the protocol.

2. **Replay state is committed before decryption.** The router advances the durable sender-index watermark after signature verification but before successful AEAD decryption/storage. A validly signed malformed/corrupted envelope with a large index can poison the watermark and cause subsequent valid lower-index messages to be discarded. Commit the index only after successful authentication/decryption, or use temporary state for early filtering.

3. **Make OTK-less bootstrap explicit.** When `oneTimePreKey` is absent, responder behavior depends on current pool availability: it may select another prekey or omit DH4. The handshake should explicitly declare the bootstrap mode; if OTK is absent, the responder must omit DH4, and if an OTK is specified it must resolve that exact key or fail.

4. **Handle malformed prekey reservation metadata defensively.** `int.parse` on persisted `reservedAt`/`inUseAt` can throw and break prekey loading. Treat malformed metadata as expired/invalid and recover safely.

5. **Fix the L3 runner safety behavior.** `.l3e2e/run.sh` says the host-Tor check is a safety gate, but it only warns and continues. Either abort when host Tor is detected or clearly document why continuing is safe. The opt-in real-Tor harness should also run in a dedicated CI job, or its results should be clearly distinguished from the normal green CI check.

I would be happy to re-review after these points are addressed. Until then, I do not recommend merging.

xmreur:
build:test

xmreur:
## PR build results

|---|---|---|
| Windows | success | [Download](https://github.com/xmreur/prysm/actions/runs/31090900070/artifacts/8963591369) |
| Linux | success | [Download](https://github.com/xmreur/prysm/actions/runs/31090900070/artifacts/8963460435) |
| macOS | success | [Download](https://github.com/xmreur/prysm/actions/runs/31090900070/artifacts/8963550853) |
| iOS | success | [Download](https://github.com/xmreur/prysm/actions/runs/31090900070/artifacts/8963640456) |
| Android | success | [Download](https://github.com/xmreur/prysm/actions/runs/31090900070/artifacts/8963682561) |

Triggered by: @xmreur
Run: https://github.com/xmreur/prysm/actions/runs/31090900070

MrBuondi:
All five points were verified against the code before anything was changed, and all five reproduce. Fixed in `2e43562`, `a2d0c7b`, `19b28a0`.

**1. High-water mark.** Confirmed. Out-of-order arrival needs no attacker: the index is allocated at send time (`group_chat_service.dart:154-157`), a failed target is queued (`:200-209`), and the retry re-sends the stored envelope with its original index (`:447-458`, driven by `sync_coordinator.dart:226`). Index 9 arriving after index 10 was acked `{status: received}`, never stored, and marked `sent` on the sender side.

Fixed by removing the ordering assumption rather than adding a window: `group_inbound_seen (groupId, senderId, msgIndex)` primary key, one row per received message, and `recordInboundIndex` is a single `INSERT OR IGNORE` — atomic, so the mutex is gone. Only the exact triple already recorded is rejected; `nextIndex` is outbound-only again. Guard: `group_message_anti_replay_test.dart`, "claim 1: a late retry of an older index is stored".

**2. Replay state committed before decryption.** Confirmed, with a wider vector than stated: signing requires the sender's key, replaying does not, and `POST /message` authenticates nobody (`PrysmServer.dart:210-224` — `senderId` comes from the body). A member relaying a captured index-10 envelope to a peer that had not received it yet made that peer drop the sender's indices 1..9, each acked as delivered. The same change closes it: with no watermark, a captured envelope can only suppress itself.

Ingress never decrypts — envelopes are stored encrypted and decrypted at render — so "commit after decryption" was not available there. With exact-duplicate detection, recording before storage is required: a soft-delete tombstone must keep winning over a re-delivery (`message_crud_dao.dart:94-113`). Guard: same file, "claim 2: a relayed captured envelope cannot censor the sender", which also asserts that a forged index-mutated envelope records nothing.

Schema consequence: the new table is created by `GroupSenderIndexStore.ensureTable`, previously reached only from `onCreate` and the `oldVersion < 7` step, so an install already at v10 would have hit `no such table`. `chat_app.db` is now version 11 with a guarded step, covered by `test/security/chat_db_v10_v11_upgrade_test.dart`, which opens a real v10 file through `DBHelper` and asserts the outbound counters survive.

**3. OTK-less bootstrap.** Confirmed, and introduced by `060ab37` on this branch: on `origin/main` `oneTimePreKeyPublic` is non-nullable and `_nextOneTimePublic` throws, so an OTK-less bundle could not exist and the responder's fallback was unreachable. Measured across an expiring reservation: 96 bytes initiator vs 128 bytes responder, then `SecretBoxAuthenticationError` and `StateError('Missing ratchet handshake')` on every later message.

Fixed without a wire field, since absence is unambiguous once the fallback is gone: `sharedSecretAsResponder` with no requested OTK derives DH1‖DH2‖DH3 and never reads the pool, and `_lookupOneTimePreKey` now takes a non-nullable key. The test that codified the fallback is inverted in the same commit; a new test asserts 96 bytes on both sides after the reservation TTL has passed.

**4. Reservation metadata.** Confirmed. The `FormatException` escaped to `key_manager.dart:37` (unlock throws instead of returning false), to `buildProfile` (`/profile` 500 for every requester) and to the handshake path; the trigger is local storage only. Two further sites throw before `int.parse` is reached — `Map<String, String>.from(cast)` and `base64Decode(entry['pub']!)` — so the fix covers the whole pool read: `int.tryParse` treats an unparseable mark as expired, and entries that are not a decodable 32-byte X25519 key are dropped and replaced by `_ensureOneTimePool`. The length check is load-bearing: a truncated base64 seed decodes fine and then throws `ArgumentError` in `newKeyPairFromSeed`, where nothing catches it. Four regression tests.

**5. L3 runner.** `.l3e2e/run.sh` now exits 1 when a host tor is running, with `PRYSM_L3E2E_ALLOW_HOST_TOR=1` as the explicit override, and the comment no longer claims the check is mirrored inside the container: the PID namespace is the containment, and the pattern-less `pkill -9 tor` at `tor_service.dart:497` fires whenever the targeted kill matches nothing, which is the normal case.

For CI I took the second of your two options. The harness needs the bundled tor binary, which is git-ignored (`.gitignore:53`), so a job running it from a checkout cannot work as things stand. The normal job now states in its own log that the real-Tor harness did not run, that it is gated behind `PRYSM_E2E=1`, and that `.l3e2e/run.sh` is required — a green check no longer implies coverage it does not have. A job that provisions tor itself would be a separate change with a run to show for it.

`flutter analyze` clean; `flutter test` 906 passed, 1 skipped (the L3 harness). Every new test was verified failing on the pre-fix code.

xmreur:
Follow-up concerns after the latest fixes:

The five original review points now look substantially addressed. I have two remaining concerns around the new `group_inbound_seen` table:

1. **Replay marking and message storage are not atomic.** The inbound path records `(groupId, senderId, msgIndex)` before storing the encrypted message. If the process crashes or the message insert fails after the seen-row insert, a retry will be treated as a duplicate and the message can be permanently lost. Please consider putting the seen-row insert and message insert in one SQLite transaction, or introduce an explicit state that distinguishes “seen and stored” from “seen but storage failed.” Tombstone protection should remain intact.

2. **The exact-seen table grows without bound.** One row is retained for every inbound group message and `resetForGroup` is the only cleanup path. Please document the intended retention/lifecycle, or add a safe pruning strategy that cannot allow old captured envelopes to be replayed successfully. If unbounded retention is intentional, please call out the storage trade-off and add an operational test/limit.

These are follow-up reliability and operational concerns rather than a reappearance of the original high-water-mark vulnerability. Please address them or explicitly track them as follow-up issues before merge.

MrBuondi:
Both addressed in `43f177d` and `92b4ea1`. One related defect found while doing it is *not* fixed here and is described at the end.

**1. Marking vs storage.** Real, and the loss path is the one you describe: the insert throws, `PrysmServer` returns 500, the sender queues and retries, and the retry hit the recorded row.

A single SQLite transaction is not available: the seen-set is in `chat_app.db` (`DBHelper`) and the messages are in `messages.db` (`lib/database/messages_database.dart`) — two databases, two connections. So this is the explicit-state option. `group_inbound_seen` now carries `resolved`, and the store exposes claim / resolve / release: the row is claimed before storage, resolved once the envelope reaches a terminal decision, released if anything in between fails. Only a resolved row is a duplicate. The whole window between claim and resolve is guarded, not just the insert — `DisappearingTimerService.resolveInboundExpiresAt` and the cache-only identity lookup are inside it too.

Recovery after a crash is not time-based. A first attempt used a 30 s claim expiry; measured against the sender's actual retry path (`tor_delivery.dart:19-21` backs off 800 ms / 2 s / 3 s, plus ~2 s of queue backoff in `group_chat_service.dart`) the first retry lands well inside that window, so it would have been refused and acked exactly as before. The claim is instead owned by the process that took it: an unresolved row whose key is not in this process's live set can only have been left by a process that died, so it is taken over immediately, whatever the retry timing. A live concurrent duplicate is still refused, so two copies POSTed under different transport ids still cannot both be stored.

Tombstone protection does not depend on this state machine: `insertInboundMessage` refuses a re-delivery whose row carries `deletedAt` (`lib/database/message_crud_dao.dart:101-113`), and the existing guard covers it. The claim resolves on that null too, so a terminal refusal leaves no unresolved row behind.

Residual, stated plainly: a crash between the message insert committing and the resolve update leaves a stored message with an unresolved claim, so a later re-delivery of that envelope is stored a second time. With the seen-set and the messages in different databases that window cannot be closed; a duplicate is the better failure than a silent loss.

**2. Unbounded growth.** Also real. The seen-set is now bounded per `(groupId, senderId)`: 256 resolved indices are retained, and pruning raises a floor in a new `group_inbound_floor` table. The floor *refuses*, it does not forget — an index below it is rejected rather than treated as new, which is what makes pruning safe against an old captured envelope.

The floor cannot be weaponised the way the original watermark was: it is raised only by deleting rows that were seen and resolved, never by the value of an incoming index, so a relayed high index still cannot push it. Unresolved claims are never pruned.

Numbers, since you asked for the trade-off: a row is ~56 bytes, so a pair costs at most ~14 KB and a 20-member group at most ~280 KB. The operational limit test resolves 306 indices and asserts the row count is pinned at exactly 256 while the floor advances. The cost is bounded out-of-order tolerance: reordering deeper than 256 received messages from the same sender is refused. The retry machinery cannot reach that depth — it re-sends the stored envelope with its original index and caps at 50 attempts — so the only reachable case is a message that exhausts every retry and is then manually resent after 256 later messages from that sender have been delivered.

**Separate defect, not fixed here.** Neither transport surfaces the receiver's outcome: `TorHttpTransport.postJson` reads the body and never inspects the status code (`lib/transport/tor_http_transport.dart:142-156`), and the WS `message` op acks before processing, logging a `>= 400` result after the fact (`lib/transport/ws_frame_router.dart:254-267`). A receiver answering 500 is therefore recorded as a delivery and no retry is produced, so the release path above only helps when the failure also breaks the connection (a crash) — not when a live receiver returns 500.

That behaviour predates this table, applies to every message type and both transports, and fixing it changes delivery semantics on the realtime path (WS would have to ack after processing, HTTP would have to treat 5xx as a failure), which needs its own tests. I did not bundle it into this PR. Say whether you want it as a blocking follow-up or a separate issue.

`flutter analyze` clean; `flutter test` 922 passed, 1 skipped. Every new test verified failing on the pre-fix code.

MrBuondi:
All 16 CodeRabbit findings were verified against the code before anything was changed: 13 are
fixed in the eight commits below, 2 are declined with the evidence that rules them out, and 1
is a real regression this PR introduced that needs a product decision from you, @xmreur —
it is the last section and the only one that should gate merge.

No finding was a fabrication: every code shape cited exists as described. But eight of them
state a consequence that does not hold, and one is worse than written; where that changed the
fix, it is called out below rather than quietly implemented as filed.

## Fixed

**`9335bbd` — the M8 redaction claim had two holes (CR7, CR9).** `logRequests()` at
`PrysmServer.dart:94-96` used the default logger; in the resolved source
(`shelf-1.4.2/lib/src/middleware/logger.dart:55-61`, `:78-82`) that builds the line from
`requestedUri.path` plus the full query and emits it with a bare `print()`, with no build-mode
gate. `GET /profile?requester=<onion>` therefore reached stdout/logcat **in release**, three
lines above the redaction this PR added at `:127-135`. It is pre-existing and untouched by the
diff, but it is the literal bypass of the M8 claim in the PR body, so it is in scope. It now
routes through `Logging`.

That only helps because `Logging` itself was fixed in the same commit: the scrub lived in
`_appendSync`, i.e. on the file sink alone, while `debugPrint` and `developer.log` received the
raw message, error and stack trace. The scrub now runs once in `_write`, before every sink, and
`_appendSync` is a plain appender. Reachable payload, since a finding without one is cosmetic:
`StateError('WebSocket not connected to $peerOnion')`
(`ws_connection_manager.dart:496-497`), logged unguarded at `file_transfer_sender.dart:242-243`
and `PrysmServer.dart:345-346` — any peer with a dead WS link triggers it.

One correction to CR9's severity: release builds persisted nothing before this commit either.
`debugPrint` is behind `kDebugMode` (`logging.dart:137`) and `developer.log` is a VM-service
no-op in AOT — checked with a `dart compile exe` probe, not assumed. The exposure was the debug
console and DevTools. CR7 is the one that reached release. The explicit `requester` redaction at
`:127-135` was kept, not removed as redundant: the central regex requires the `.onion` suffix,
and the query value is not guaranteed to carry it.

**`600270d` — the corrupt-store guarantee did not cover a corrupt store (CR4, CR3).** The filter
added in `a2d0c7b` documents at `prekey_bundle.dart:174-179` that a corrupted or foreign store
"must not brick unlock or /profile". It runs after `jsonDecode(raw)` at `:169`, which was
unguarded, so a non-JSON blob threw straight past it. Confirmed blast radius: `key_manager.dart:37`
and `:60` call `loadStored` with no catch and neither `unlock_controller.dart:71` nor
`onboarding_screen.dart:193` catches either, so unlock hard-fails; and `buildProfile`
(`inbound_message_router.dart:122`) propagates to `PrysmServer.dart:173-176`, making `/profile`
a 500 for every peer. The decode now degrades to an empty pool and `_ensureOneTimePool` refills.
The behaviour is pre-existing (`da92dc2:prekey_bundle.dart:119`); the over-broad guarantee is
ours, which is why it is fixed here rather than deferred.

CR3 in the same commit: a negative `Duration` is always `< _reservationTtl`, the refill paths are
length-based (`:205-213`, `:303-320`) so a pinned entry is never replaced, and a future `inUseAt`
makes `_lookupOneTimePreKey` (`:267-270`) refuse that exact key — surfacing as a `StateError` and
a `POST /message` 500. Trigger is local only (restored backup on a clock-behind device, or a
backwards clock step). `_isUnexpired` now treats a future mark as expired.

**`6ddf869` — dead catch clauses removed (CN7).** Confirmed there is no later broader `catch` for
the two `rethrow`-only clauses to shadow (order: `on FormatException` → `on StateError` →
`finally`), so deleting them is a no-op, verified by execution.

**`ef016e5` — the resolve is now contained (CR6).** Accepted, with the consequence corrected: the
retry does **not** take the unresolved row over. The key is still in this process's `_liveClaims`,
so the second claim is refused at `group_sender_index_store.dart:178-181`; takeover requires a
restart (`:183-186`). And on the HTTP path the sender never sees the 500 at all, because of the
transport gap already declared in the previous round. The real residue was a delivered message
answered with a 500 and a row left unresolved forever — worth fixing on its own, and worth fixing
*before* that transport gap is closed, since closing it would make the duplicate reachable.

**`6ec407a` — orphaned claims are pruned (CN5).** Small, and taken because it contradicts the
retention bound stated in the last round: `_prune` deleted only `resolved = 1` rows, so an orphan
left by a dead process stayed forever once the floor passed its index. Rows below the floor are
now deleted regardless of `resolved` — the floor already refuses those indices, so they carry no
information. The floor is still raised only by pruning seen-and-resolved rows, never by an
incoming index.

**`9914045` — `101` is parsed as a status-code token (CR2).** Accepted as spec compliance, not as
a security fix, and the tests say so: the exact `Sec-WebSocket-Accept` check added by this PR is
mandatory, so a peer answering `1010` still has to produce a byte-correct accept header. The
substring test is pre-existing (`da92dc2:tor_socks_websocket.dart:24`).

**`1597fdc` — harness runner and image (CR1, CN2, CN1).** CR1's causal chain does not hold and is
not what was fixed: `run.sh` overrides only `PUB_CACHE`, so the container `HOME` stays
`/home/ubuntu` (`Containerfile:42`), and a bind mount carries the host directory's ownership, so
the write succeeds today — verified by running the exact mount shape under docker. What is real:
if the host `~/.pub-cache` does not exist the daemon creates it `root:root` and uid 1000 cannot
write, and a host user not on uid 1000 fails the same way. Both are fixed with the proposed mount
plus a `mkdir -p`. CN2: the digest is genuine — it matches Google's `releases_linux.json` entry
for 3.44.8 stable linux byte for byte — and `sha256sum -c` now runs between the curl and the
extraction. CN1: the arithmetic is wrong (the retry loop curls the **victim** onion at `:319`, so
an unreachable attacker costs ~8 minutes, not ~20, and the 40-minute `@Timeout` is never at risk),
but a precondition belongs before the expensive loop regardless, so it moved.

**`1cbad6d` — test guards (CN4, CN3).** CN4 is worse than filed: `claimedAt` is not a column in
any schema version, so `isNot(containsAll(['claimedAt', 'resolved']))` was satisfied by its
permanent absence and the guard could never fail — including against the one wrong fixture it
exists to catch, which would have made the whole upgrade test vacuous. Split into two assertions
and proven to bite by temporarily adding `resolved` to the fixture. CN3 is convention hardening,
not a bug fix: the stated pollution cannot fire today (the router always resolves or releases its
claims, and each case gets a fresh in-memory DB), but the two suites now match the two that
already reset.

## Declined, with the evidence

**CR5 — a throw inside `finally` replacing the original exception.** The Dart semantics are real
and were confirmed by execution, but the trigger cannot fire: `CryptoKeyStore.write` swallows
every secure-storage failure (`key_store.dart:123-127`), the only other throw source inside that
`finally` was CR4's `jsonDecode`, which the body already reaches first (`ratchet_service.dart:216`
vs `:226`), and no caller branches on the exception type. After `600270d` the guard would be
unreachable code — four lines below the two unreachable clauses CN7 asks us to delete.

**CN6 — the unguarded `rows.first` after a failed insert.** The code shape is real, the window is
not: the whole block runs inside the single static `_mutex` (`group_sender_index_store.dart:139`)
and every other writer takes it (`:199`, `:292`, `:315`); no second process writes `chat_app.db`
in production; and `Mutex.protect` releases in a `finally`, so even a hypothetical `StateError`
is caught by the callers and self-heals on the next claim.

## Needs a decision — CR8, and it is ours

This is the only finding that blocks merge, and CodeRabbit is right about it. On `da92dc2`,
`_resolveSenderIdentity` resolved over Tor on a cache miss
(`da92dc2:group_service.dart:533-541`, `resolvePeerIdentityForIngress(..., fetchOverTor:)`).
Commit `203b1f0` in this PR made it DB-only (`group_service.dart:549-551`) to close the M2 oracle
on the group path. The caller drops on `null` **before** decryption (`:481-489`), and the identity
is not optional there — `GroupCryptoV2.decryptControlPayload` (`:493-497`) authenticates against
it. `DBHelper.ensureUserExist` (`db_helper.dart:230-247`, called at
`inbound_message_router.dart:334` before the handler at `:342-346`) inserts `publicKeyPem: null`
and no `identityJson`, so it creates a row that resolves to `null`.

Proven end-to-end against the real router: a correctly signed group invite from someone the
invitee has never added is dropped silently — the sender gets 200 `{status: received}`, the
invitee sees nothing, no retry. The same drop hits key-rotate, member-removed, profile-update and
disappearing-timer from any roster member whose row has no identity material, and invite roster
sync adds no user rows (`group_service.dart:635-646`).

This is not a hardening gap, it is the M2 fix's own cost: the call that was removed *is* the
outbound `GET /profile` M2 exists to suppress. The three ways out are not equivalent, and I am
not picking one at intuition:

1. **Accept it** — first contact must precede a group invite. No code, but the drop needs a UI
   surface (today it is invisible) and a line in the PR body.
2. **Carry the inviter's identity key in the invite and TOFU-persist it after verifying the
   payload against it.** No network fetch, no oracle — but the key is bound to nothing but
   itself, since the onion address is a separate Tor service key, so an unauthenticated peer
   could seed a users row and pre-empt the real peer.
3. **Restore the fetch for `groupInviteType` only.** Smallest diff, and it reopens M2 on the one
   path an unknown peer can reach unauthenticated.

Worth noting for option 3: the 1:1 path still fetches pre-authentication
(`direct_message_auth.dart` via `PrysmServer.dart:73`), which was declared as a known residual in
the earlier round — so "no pre-auth outbound fetch" is already true only of the group paths.

Tell me which one you want and I will implement it here; otherwise I will open it as a separate
issue and this PR ships with the drop documented.

`flutter analyze` clean; `flutter test` **930 passed, 1 skipped** (was 922). Every new test was
verified failing on the pre-fix code, including the two that only bite when a fixture is wrong.

xmreur:
I recommend **Option 1: accept the DB-only identity requirement**, but make the failure visible rather than silently dropping the invite.

Please keep the current behavior of not fetching over Tor for an unknown sender. Option 2 (TOFU-persisting the inviter key) lets an attacker seed a key for an onion and pre-empt the real peer, while Option 3 (restoring the fetch for invites) reopens the M2 profile-fetch oracle.

For the UX/product follow-up, store an unverifiable first-contact invite as pending, do not process it or fetch the sender profile, and show the user that identity verification/contact addition is required. Once the sender is locally trusted, reprocess the pending invite. If that UX is too large for this PR, document the current silent-drop behavior and track the pending-invite flow separately.

The remaining transport status-code/delivery-outcome defect should also be tracked as a separate follow-up issue; I would not reopen the M2 path to solve it here.

MrBuondi:
Second CodeRabbit pass, five points. Three fixed in `b4b87b2`, `75628bf`, `6fb66ce`; one is a
correction to CodeRabbit's own suggestion, proven from the SDK source; one is the open CR8
decision restated, still waiting on @xmreur.

## NR2 — right invariant, wrong cause, real bug (`b4b87b2`)

> An active low-index claim can still be storing when another delivery advances `newFloor`.
> This delete removes that claim.

The delete is not what loses the message. `claimInboundIndex` checks the floor at
`group_sender_index_store.dart:150-153`, **before** any insert or row lookup, so a retry below
the floor is refused whether or not the row still exists. Removing the delete added in
`6ec407a` would not save that message; the loss arrived with the floor itself in `92b4ea1`.

But the invariant you name is genuinely missing, and it is the actual defect: **the floor could
be raised past an index whose claim was still in flight.** After that the claim can never be
released and retried, so a store failure on that message becomes a permanent loss. Fixed by
clamping, not by dropping the prune: `_prune` now walks the unresolved rows below the candidate
floor and pulls the floor down to the lowest one **this process holds live**, so a live claim
always sits at or above the floor and stays claimable (the floor refuses `index < prunedBelow`,
so a claim exactly at the floor is still admitted).

Keying the clamp on process ownership rather than on `resolved` is what keeps both properties:
an unresolved row this process does not own is an orphan a dead process left between claim and
resolve, it is already recovered by the ownership takeover at `:183-186`, and it must still be
pruned or the retention bound this branch promised in the last round is false again. That is the
same rule the takeover path already uses, so there is no second concept.

Both of your test cases are in, plus the one that guards against overcorrecting:
- a live low claim holds the floor down, is released, and its retry claims successfully — red on
  the pre-fix code with `prunedBelow` at 51;
- an orphan below the floor is still pruned while a live claim clamps the floor — red both on
  the pre-fix code and against a naive "protect every unresolved row" variant, which is exactly
  the regression that variant would have caused.

## NR3 — `developer.log` name field (`75628bf`)

Correct, and consistent with what `9335bbd` was for: the scrub is supposed to have one home and
cover every value that reaches a sink, and `name:` is one. Fixed as suggested. Honest scope: no
call site passes anything but a literal class name, so this is defence in depth, not a live
leak, and there is no red to show for it — the field is only observable through an attached VM
service, which a `flutter test` isolate does not expose.

## NR1 — host uid mapping (`6fb66ce`)

Accepted, and it is the residual the last round left open: the mount fix in `1597fdc` closes the
missing-directory case but not a host that is not uid 1000, and `Containerfile:39-40` was still
asserting that assumption in a comment. The image now remaps its `ubuntu` user from
`HOST_UID`/`HOST_GID` build args that `run.sh` passes from `id -u`/`id -g`, with the remap placed
after the Flutter install so changing the args does not invalidate the ~1 GB download layer, and
the `chown -R /opt/flutter` moved after the remap so the SDK does not keep the old uid. Running
as root aborts early rather than producing a subtly broken image.

Verified by execution rather than by reasoning, on throwaway images so no large layer was
duplicated: a mount owned `2000:2000` is **not writable** by the pre-fix image (`uid=1000`,
exit 1) and **is writable** by the post-fix image built with `HOST_UID=2000` (`uid=2000`,
exit 0); the `/opt/flutter` marker follows the new uid, proving the chown ordering; default args
are a no-op; and the real image still builds with the curl layer cached.

## NR5 — the host-tor gate (`6fb66ce`)

You are right that the gate was in the wrong place, and the script's own comment admitted it
("it protects the host case") while the script only ever takes the container path. It was not
removed, though — it was moved to where the danger actually is.

Deleting it outright would undo what @xmreur asked for in the first round, and the hazard is
real: `TorManager`'s pattern-less `pkill -9 tor` (`tor_service.dart:497`) fires whenever the
targeted kill matches nothing, which is the normal case. What is safe is the *container* path,
because of the PID namespace — not the harness. So the gate now lives in
`headless_inbound_e2e_test.dart`'s `setUpAll`, as its first statement, before any daemon starts:
on the host with a live tor it fails the test with the same message and the same
`PRYSM_L3E2E_ALLOW_HOST_TOR=1` override; inside the container it passes, because the image starts
no tor of its own. `run.sh` keeps only the comment explaining why its path is safe. `pgrep` is
present in the image (`procps`, verified against the rebuilt image: procps-ng 4.0.4).

## NR4 — declined, and the suggestion does not hold on this SDK

> Wrap the loopback server and client in `HttpOverrides.runZoned`; its default override provides
> the real `HttpClient` without changing global state.

The concern is right — `HttpOverrides.global = null` mutates root-zone state and `HttpOverrides`
has no getter, so it cannot be restored — but the proposed remedy does not work here.
Zero-argument `HttpOverrides.runZoned` builds an `_HttpOverridesScope(null, null)` whose
`_previous` is `HttpOverrides.current`, which under `flutter_test` is the binding's
`_MockHttpOverrides` (installed by `TestWidgetsFlutterBinding`, `_binding_io.dart:26-27`), and
`createHttpClient` delegates straight to it. Checked empirically as well: with the test wrapped
in `runZoned` the request never reaches the loopback server and comes back 400 from the mock.

The global clearing therefore stays, with a comment recording why, and the test now asserts
`statusCode == 200` — the mock answers 400, so that assertion is what proves the request really
crossed a socket instead of silently passing against a mock.

## Still open — the unknown-inviter policy

Your outside-diff comment on `group_profile_fetch_gate_test.dart:275-293` is the same finding as
the previous round's CR8, and it is the one thing that should gate merge. It is a regression this
PR introduced (`203b1f0` replaced the Tor fallback in `_resolveSenderIdentity` with a DB-only
lookup to close M2 on the group path), it is proven end-to-end, and the three exits you list are
the same three I put to @xmreur — they are not equivalent, so it is his call, not a mechanical
fix. The test you ask for lands with whichever policy he picks.

`flutter analyze` clean; `flutter test` **933 passed, 1 skipped**.

MrBuondi:
Option 1 it is. Taken exactly as you scoped it, including the fallback: the pending-invite UX is
tracked separately rather than bolted on here, for a reason worth stating.

**The security boundary is unchanged.** `_resolveSenderIdentity` stays DB-only. No Tor fetch for
an unknown sender, on any group control path.

**`f94e2a8` — the accepted behaviour is now documented and pinned.** A comment at the drop branch
(`group_service.dart:481-489`) records that the drop is deliberate, that the identity is
*required* rather than merely checked — `GroupCryptoV2.decryptControlPayload` authenticates
against it, so the payload cannot be read without it — and that restoring a network resolve there
would reopen the M2 oracle, so it must not be "fixed" that way. Two `POLICY:` tests in
`test/security/group_profile_fetch_gate_test.dart` now hold the line: a correctly built
first-contact invite is acked, not applied, no group or member rows, and no profile fetch; the
same invite with the inviter's identity stored locally IS applied. The pair matters — without the
positive case the first test would also pass if invites were broken outright.

The first test is red on the pre-PR behaviour (`Expected: empty, Actual: ['inviter.onion']` — the
old fallback resolved the inviter over the network and applied the invite). Proving that needed a
fake HTTP layer in the test: under `flutter_test` every outbound request gets an instant mocked
400, so a reintroduced fetch would fail silently and leave the test green. The fake is inert
while the policy holds.

**#129 — pending first-contact invites.** Your flow, recorded as specified: store as pending, do
not process, do not fetch, tell the user contact addition is required, reprocess once the sender
is locally trusted. One design constraint is written into the issue because it is easy to miss:
that store is written by **unauthenticated** inbound traffic — any peer can POST an invite
envelope naming any `senderId`, and `POST /message` authenticates nobody — so an unbounded
pending table is a remote write primitive and a storage-exhaustion vector. Same class as the
`group_inbound_seen` growth you caught in the second round. It needs per-sender and global bounds
designed in, which is most of why it does not belong in this PR.

**#130 — transport delivery outcome.** Filed, and not touched here. Noted in the issue that it
interacts with the claim/release path from this PR: the release only helps when the sender learns
the delivery failed, which today it does not on the HTTP path. Closing #130 makes that residue
reachable, so whoever picks it up should read the two together.

**PR body updated** with the accepted behaviour, both issue links, and the honest qualifier that
"no pre-auth outbound fetch" holds for the group paths — the 1:1 path still resolves an unknown
sender before verification (`direct_message_auth.dart` via `PrysmServer.dart:73`), which is the
separate decision flagged earlier.

`flutter analyze` clean; `flutter test` **935 passed, 1 skipped**. Nothing open from my side.

xmreur (approved):

xmreur:
build:test

xmreur:
## PR build results

|---|---|---|
| Windows | success | [Download](https://github.com/xmreur/prysm/actions/runs/31111616721/artifacts/8972183430) |
| Linux | success | [Download](https://github.com/xmreur/prysm/actions/runs/31111616721/artifacts/8972025878) |
| macOS | success | [Download](https://github.com/xmreur/prysm/actions/runs/31111616721/artifacts/8972141941) |
| iOS | success | [Download](https://github.com/xmreur/prysm/actions/runs/31111616721/artifacts/8972188411) |
| Android | success | [Download](https://github.com/xmreur/prysm/actions/runs/31111616721/artifacts/8972348212) |

Triggered by: @xmreur
Run: https://github.com/xmreur/prysm/actions/runs/31111616721

PR #133, opened 2026-08-07
Group invite mode (1/3): setting and bounded pending-invite store

MrBuondi:
## What

First of three stacked PRs implementing the user-selectable group invite mode. This one adds only the data layer — no behaviour change to the inbound path yet.

- `GroupInviteMode` (`lib/models/group_invite_mode.dart`): two-value enum with `label` and `description`, so both choices can be explained in the UI without a tap.
- `Settings.groupInviteMode` + getter/setter on `SettingsService`. Default is `holdAsRequest`.
- `GroupPendingInviteStore` (`lib/util/group_pending_invite_store.dart`): bounded store for first-contact invites, plus `chat_app.db` `version: 14`.

## Why the store is bounded the way it is

This table is written by **unauthenticated** traffic: `POST /message` authenticates nobody, and the only rate limit is keyed on the sender-claimed `senderId` (`lib/server/PrysmServer.dart:223-228`). Without bounds it is a remote write primitive, so all four are load-bearing:

| Bound | Value | Why this and not something else |
|---|---|---|
| Per sender | 1 row (primary key); a newer invite replaces the older | One onion occupies one slot, not N |
| Global | 20 senders; **at capacity the new row is refused**, no eviction | Evicting the oldest would let an attacker delete a genuine request. Refusing degrades to today's drop, which is already the accepted worst case |
| Wire size | ≤ 64 KiB per envelope | A real `control-wrap-2` envelope is a few KB. The general inbound cap allows a 96 MiB body, so 20 rows at that size would pin ~1.9 GB of attacker-chosen bytes for the whole retention window |
| Retention | 7 days, pruned on the write path | Self-healing: a table filled by an attacker frees itself with no user action |

The row is the raw `control-wrap-2` envelope, stored **opaque**. It is never decrypted while pending: the payload is attacker-controlled and unauthenticated, so no field of it may reach the UI or the database in parsed form.

## Schema

`chat_app.db` goes to `version: 14`. `ensureTable` is called both from `_createCryptoTables` and from a **guarded ladder step**, because a table added only in `onCreate` would be missing on every installation already at v13 and would fail with `no such table` on the first group message.

## Verification

- `flutter analyze`: 0 issues.
- `flutter test`: all green on this branch alone.
- Upgrade path exercised on a real file, not in memory: `test/security/chat_db_v13_v14_upgrade_test.dart`.
- Bounds have tests that fail against the pre-fix code, including the exactly-at-the-bound case (`test/security/group_pending_invite_store_test.dart`).

## Review order

Stacked. Merge this first, then `invite-mode/policy`, then `invite-mode/ui`.

## Summary by CodeRabbit

* **New Features**
  * Added configurable handling for group invites from unknown senders.
  * Users can hold invites as pending requests or accept invites only from contacts.
  * Added a pending invite requests screen with accept and discard actions.
  * Added privacy settings, navigation, and pending-request counts.
  * Pending requests are securely retained with size limits and automatic expiry.
* **Bug Fixes**
  * Preserved existing handling for other message types.
* **Tests**
  * Added coverage for settings, persistence, upgrades, storage, and invite workflows.

### DIFF
--- a/docs/superpowers/plans/2026-08-07-group-invite-mode.md
+++ b/docs/superpowers/plans/2026-08-07-group-invite-mode.md
@@ -1,0 +1,1565 @@
+# Group Invite Mode Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Let the user choose, in Privacy Settings, whether a group invite from someone not in
+their contacts is discarded on arrival (today's merged behaviour) or held as a pending request
+they can accept — with both modes explained by a subtitle at the point of choice.
+
+**Architecture:** A two-value enum setting read synchronously by `GroupService` at the existing
+unknown-sender drop branch. In hold mode the raw, still-encrypted control envelope is written to
+a new bounded table in `chat_app.db` and nothing else happens: no decryption, no outbound
+traffic, same `200 received` ack. The user accepts from a dedicated screen, which runs the
+existing user-initiated contact-add (the only fetch in the whole flow) and then replays the
+stored envelope through the normal authenticated path.
+
+**Tech Stack:** Flutter/Dart, `sqflite` (+ `sqflite_common_ffi` in tests), `mutex`,
+`shared_preferences`. No new dependencies.
+
+**Spec:** `docs/superpowers/specs/2026-08-07-group-invite-mode-design.md`
+
+## Global Constraints
+
+- No new dependencies, in neither `dependencies` nor `dev_dependencies`.
+- `flutter analyze` must report **0 issues** and `flutter test` must stay green before every
+  commit. Baseline on this branch: `935 passed, 1 skipped`.
+- User-facing strings are hardcoded English (the app has no i18n layer). Comments and commit
+  messages in English; conventional one-line commit subjects.
+- Reuse existing patterns; a second convention beside an existing one is prohibited. The
+  patterns this plan follows are named per task with `path:line`.
+- Tests: hand-written fakes, `sqflite_common_ffi` in-memory databases, the
+  `setDatabaseForTest` / `resetForTest` seams. Every test must be seen failing before the
+  implementation makes it pass.
+- **The security boundary does not move.** `_resolveSenderIdentity` stays DB-only; nothing in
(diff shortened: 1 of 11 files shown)


MrBuondi:
All three verified against the code before touching anything. All three real, all three fixed — no false positives this round.

**CR1 — `take` contract in the spec.** Correct: the spec listed `Future<Map<String, Object?>?>`, the implementation is `Future<String?>`. Fixed in `dc5550a`, together with the retention row of the bounds table, which claimed pruning happened only "on the write path" — no longer true once `take` prunes too.

**CR2 — measure the bound in UTF-8 bytes.** Correct *and reachable*. The wire is `data['message']` straight off `POST /message`, and the only validation on it is `data['message'] is String` (`inbound_message_router.dart:854`), so the doc comment's premise — "the envelope is base64 and JSON, i.e. ASCII" — was an unenforced assumption about attacker-controlled input. Worst case is three UTF-8 bytes per UTF-16 code unit: 192 KiB per row against a declared 64 KiB, so ~3.84 MiB pinned instead of ~1.28 MiB. Bounded, but wrong. Fixed in `fe6fe3e`, with the constant renamed `maxPendingWireBytes` so the name cannot drift from the unit again.

One deliberate deviation from the suggested patch: `utf8.encode(wire).length` **on its own** would put a DoS inside the guard meant to prevent one. The general inbound cap allows a 96 MiB body, and encoding it just to measure it allocates ~288 MiB. So the cheap `String.length` test stays as a first short-circuit — UTF-8 is never *fewer* bytes than UTF-16 code units, so anything over the bound in units is over it in bytes — and the encode only ever runs on a payload already known to be ≤ 64 Ki code units.

New test `the bound counts UTF-8 bytes, not UTF-16 code units` uses `'€'` (1 code unit, 3 bytes) on both sides of the byte boundary. Proven mordant: it fails with the byte check removed, passes with it in place.

**CR3 — prune before `take`.** Correct, and worth being precise about why it read that way. The prune does exist — it just arrived one PR later in this stack, in #134 (`feat(group): promote held invites once the sender becomes known`). At *this* PR the store has no production caller at all: `git grep GroupPendingInviteStore -- lib` on this branch returns only `ensureTable` from `db_helper`, so the gap was not reachable here.

It was still in the wrong PR. Pruning on a read path is a store concern, i.e. this PR's topic, not the promoter's. Moved here in `a8d1e51`, with a test that a row past the retention window makes `take` return null.

Gate: `flutter analyze` 0 issues, `flutter test` green on this branch alone.

PR #136, opened 2026-08-07
Live-app-testing toolkit: container, prysmlab CLI, and skill

MrBuondi:
## What

Tooling that lets a developer — or an agent — drive the **real** Prysm app and read its widget tree back, instead of only running the test suite. Independent of the group-invite-mode PRs; based directly on `main`.

| Path | Role |
|---|---|
| `tool/live/Containerfile` | `prysm-lab` image: ubuntu 24.04 + Flutter 3.44.8 + GTK + Xvfb |
| `tool/live/labroot.sh` | root-side bootstrap: system D-Bus + `upowerd` |
| `tool/live/labd.sh` | in-container start: Xvfb, keyring, `flutter pub get`, `flutter run` |
| `tool/live/prysmlab` | Python 3 **stdlib-only** CLI, 23 commands, single entry point |
| `.agents/skills/live-app-testing/SKILL.md` | when to use it, the safety gate, and 12 traps with symptom and cause |

Three commands to a running, onboarded app:

```sh
./tool/live/prysmlab doctor
./tool/live/prysmlab up
./tool/live/prysmlab onboard --pin 112233
# ... work ...
./tool/live/prysmlab down --purge
```

Everything that is not lifecycle forwards itself into the container, so `prysmlab screen` typed on the host does a `docker exec` on itself.

## Why a container and not a sandbox directory

Not a preference — two mechanisms in the app make it mandatory:

- `TorManager._killOrphanTorForce` (`lib/util/tor_service.dart:497`) runs `pkill -9 tor` **with no pattern**. Only the PID namespace contains that; outside a container it kills the developer's system tor.
- The app's server binds `127.0.0.1:12345`. Only the network namespace stops it colliding with a real Prysm on the same machine.

The first five layers are byte-identical to `.l3e2e/Containerfile` so Docker reuses the cache and the ~1 GB Flutter SDK download does not repeat. **If `.l3e2e/Containerfile` changes, keep them aligned** or the build goes back to ~20 minutes.

## Design notes worth knowing before changing it

- **`evaluate` context is `package:flutter/src/widgets/binding.dart`**, not `package:prysm/main.dart`. That library already imports foundation, gestures, rendering, scheduler and services, so `WidgetsBinding`, `Element`, `RenderBox`, `Offset`, `PointerDownEvent` and `HitTestResult` are all in scope at once.
- **Taps hit-test before firing.** `localToGlobal` returns coordinates for off-viewport widgets too, so a naive driver reports success for a tap at y=1417 on a 720px-high window. `NOTFOUND` (absent) and `OCCLUDED` (covered) are now distinguishable, and autoscroll is the default.
- **`onboard` reacts to what is on screen** instead of replaying a fixed key sequence, which is why it handles the first-run wizard, the PIN unlock and "Reset local data and continue" with one code path.
- **Stdlib only**, including a hand-written RFC 6455 client (~100 lines). Zero new dependencies for the project.

## Three claims in the previous manual runbook that turned out to be false

| Claimed | Actual |
|---|---|
| First start needs a human click on the keyring dialog | Launcher bug: `printf ''` sends EOF with no newline, so the daemon never creates `login.keyring`. With `printf '\n'` it is fully headless — two `dbus-run-session` store+lookup cycles took 0.06 s total |
| Every restart begins again from "Local data error" | Same bug. With a real keyring the identity persists: `restart` lands on `UnlockScreen` in 5.4 s |
| `evaluate` only fails with a precompiled bundle | It also fails under `flutter run` when you talk HTTP instead of WebSocket — `compileExpression` is registered on the WS session. Same error text, different cause |

Two environment faults with an identical mute symptom ("app compiles, window opens, `debugDumpApp` says `<no tree currently mounted>`") are fixed in the image: missing `xdg-user-dirs` → `MissingPlatformDirectoryException`; missing system D-Bus → `battery_plus` hangs `AppBootstrap.initializeServices` and `runApp` is never reached.

## Verification

Full cycle run from a completely empty state — image deleted, only the files in this PR, nothing installed warm:

| Step | Observed | Time |
|---|---|---|
| `up` (no image) | rebuild, stage, container, system bus, Xvfb, keyring, Linux bundle, `OnboardingScreen` rendered | ~100 s |
| `onboard --pin 112233` | `[9] home reached: ['HomeScreen']` | 22 s |
| `peer public` / `peer onion` | real identity JSON; real `.onion`, Tor bootstrapped in-container | — |
| `count` → `tap "Chat with myself"` → `count` | `1` → `2` | — |
| `type "hello-lab"` | `TYPED n=1 now=hello-lab` | — |
| `tap-widget --contains "Settings"` → `pop` | opens `SettingsScreen`, then `CANNOT-POP` | — |
| `restart` | `UnlockScreen`, identity intact | 5.4 s |
| `down --purge` | verification block reports everything removed | 2 s |

This tooling found a real defect: the pending-invite counters not refreshing, fixed in the `invite-mode/ui` PR of the group-invite-mode series.

## What this does not prove

Stated plainly because it is easy to over-claim:

- **Not graphical rendering.** `shot` writes a real PNG from the Xvfb framebuffer, but nothing here asserts what is in it.
- **Not the compositor's input path.** Taps are injected via `WidgetsBinding.handlePointerEvent`, which is below the platform input layer.
- **Not CI.** It needs Docker and a real Tor bootstrap.

## Hygiene

No user data was read or modified; the real identity was never unlocked and the real app was never launched — everything happened inside the container. The PIN `112233` is the CLI default for a throwaway identity, not a secret.

## Summary by CodeRabbit

* **New Features**
  * Added an isolated environment for validating the Prysm desktop app end to end.
  * Added tools to launch, monitor, and control the app, including navigation, tapping, scrolling, text entry, screenshots, logs, and status checks.
  * Added automated onboarding and peer-request workflows.
  * Added documentation describing supported testing workflows, safety guidelines, troubleshooting, and cleanup.

* **Chores**
  * Added container and environment setup for headless Linux desktop testing.

### DIFF
--- a/.agents/skills/live-app-testing/SKILL.md
+++ b/.agents/skills/live-app-testing/SKILL.md
@@ -1,0 +1,220 @@
+---
+name: live-app-testing
+description: "Use when you need to prove UI<->state integration on the real, running Prysm desktop app: screens render what the stores contain, taps land, routes open and close. Proves the app's own wiring end to end. It does NOT prove pixels, does NOT prove the compositor's input path, and does NOT run in CI."
+---
+
+# Live App Testing
+
+Drive the real Prysm Linux desktop app with `tool/live/prysmlab` — observation (Dart VM
+Service widget trees), control (in-app pointer events via `evaluate`), and data injection
+(HTTP against the app's loopback server). The app runs in a throwaway container under Xvfb
+with a private D-Bus session and keyring.
+
+## When to use it — and when not
+
+Use it when the defect lives in the integration between UI and state: a screen that does not
+show what the DB holds, a tap that does not reach its handler, a route that does not open or
+close, a store whose change the UI never reflects. That is where this method found real bugs.
+
+Do NOT use it for:
+
+- **Pixels.** The dump is text: what is in the tree, not how it looks. Alignment, contrast,
+  truncation, overlap are invisible here. `prysmlab shot` writes a PNG of the Xvfb framebuffer
+  (verified: 1440x900), but producing an image is not reading one: with no vision-capable tool
+  available, the PNG proves nothing on its own.
+- **The compositor's input path.** `prysmlab tap` delivers `PointerDown/UpEvent` through
+  Flutter's own hit test and gesture recognizers. It proves the app's wiring, not that a real
+  click (window focus, display scaling, layered windows) arrives the same way.
+- **Regression.** It is a repeatable manual proof, not a test: it does not run in CI and does
+  not fail on its own. Turn findings into real tests afterwards.
+
+## Safety — the container is not decoration
+
+`TorManager._killOrphanTorForce` (`lib/util/tor_service.dart:497`) runs a bare `pkill -9 tor`
+when its targeted pkill misses — the normal case. The container's PID namespace is the only
+thing between that and the host's tor. The app's server binds `127.0.0.1:12345`; the network
+namespace is the only thing between the injection channel and — or blocked by — the user's
+real Prysm. Rules:
+
+- Never run the app outside the container.
+- Never unlock the user's real identity. If "Enter Passcode" appears and you did not create
+  the lab identity yourself, isolation broke: stop.
+- Use a throwaway PIN (`prysmlab onboard --pin 112233`). The lab identity lives in the
+  container's writable layer — its keyring is `/home/ubuntu/.local/share/keyrings`, not the
+  pub-cache volume — so it dies with the container, and `up --fresh` or any `down` ends it.
+- The repo is never written to: `up` stages a copy into `/tmp/prysm-lab/work` and mounts that;
+  `tool/live` is mounted read-only.
+
+## Workflow
+
+```sh
+prysmlab doctor           # preflight: docker, image, rsync, tor binary; shows host tor/port are irrelevant
+prysmlab up               # stage repo, start container, start app; waits for the VM Service line
+prysmlab onboard --pin 112233   # reaches home from wherever the app is: first-run wizard,
+                                # passcode unlock, or local-data reset, by reading the screen
+# observe and act, in a loop:
+prysmlab screen           # what is on screen, from the widget tree
+prysmlab tap "Settings"   # act
+prysmlab down --purge     # teardown is part of the method — always
+```
+
+`up` prints progress; the first run compiles the Linux bundle (minutes). `up --fresh` destroys
+the container first (loses the test identity) and re-stages; `prysmlab sync` re-stages without
+relaunching (a running app keeps the old code); `prysmlab restart [--sync]` relaunches the app
+in the running container — the way to pick up code changes (hot restart over stdin is not an
+option: sending `R` to `flutter run` makes it exit; `--sync` re-stages first). All commands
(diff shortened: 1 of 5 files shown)


MrBuondi:
All five verified against the code before touching anything. All five real, all five fixed — no false positives. Two of them I reproduced in a container rather than reasoning about them.

**CR3 — the uid/gid remap fails when the host id already exists.** Correct, and reproduced. In `ubuntu:24.04`, `getent group 20 100` returns `dialout:x:20:ubuntu` and `users:x:100:`, and `groupmod -g 20 ubuntu` answers `groupmod: GID '20' already exists`. Inside a `&&` chain that aborts the whole build, so any host user on gid 20 or 100 — the macOS `staff` default among them — could never build the image. Fixed with `--non-unique` on both `groupmod` and `usermod` in `b673de4`; verified in the same container that `chown -R ubuntu:ubuntu` still resolves by name afterwards, which is all this step needs.

**CR4 — the stdin holder outlives the app.** Correct, and the mechanism is exactly as described: `sleep` never writes, so it never takes `SIGPIPE`, so the shell kept waiting for it after `flutter run` exited — holding `dbus-run-session` and its `gnome-keyring-daemon` alive. `restart` kills only `flutter_tools.snapshot run` (`prysmlab:609`), so every restart added one stale session bus and one stale keyring daemon.

Fixed in `cf2f0db`, but not with a FIFO *plus a writer* — with a FIFO opened **read-write**:

```sh
flutter run … <> "$FIFO"
```

The fd is its own writer, so it never signals EOF and there is no holder process to leak in the first place. That deletes a process instead of adding one. Verified on the real `/bin/sh` of `ubuntu:24.04` (dash) that `<>` on a FIFO neither blocks on open nor yields EOF. `flutter run`'s exit status is preserved explicitly, since the trailing `rm` would otherwise mask it.

**CR5 — fail when UPower is unavailable.** Correct on both halves, and the second half is the sharper one: the wait lived *inside* `if ! pgrep -x upowerd`, so it was skipped precisely when a stale daemon that does not own the name is the problem. And the loop had no verdict after it, so on failure the script still printed `PRYSMLAB_SYSTEM_BUS_UP` and `up` went on to start the app.

Fixed in `faf508c`: the wait moved out of the `if` and runs unconditionally, waits for the bus **name** rather than for a process, and exits non-zero with `PRYSMLAB_UPOWER_FAILED`. `prysmlab up` now recognises that marker and reports the actual cause instead of the generic system-D-Bus message — the alternative symptom was a bare timeout three layers away from the cause, which is the exact failure mode this file exists to prevent.

**CR1, CR2 — `afterward`, `self-chat`.** Both correct, fixed in `3b57882`.

Skill rule 10 updated in `b95e17c` to state the new hard gate, so the document no longer describes a leniency that has been removed.

Gate: `flutter analyze` 0 issues, `flutter test` green (this PR adds no Dart, so it matches `main`'s baseline). Both shell scripts syntax-checked with `sh -n` under dash inside `ubuntu:24.04`; `prysmlab` parsed with `ast.parse`. The first eight layer-forming instructions remain byte-identical to `.l3e2e/Containerfile`, so the ~1 GB Flutter SDK layer stays shared and cached.

PR #137, opened 2026-08-07
Group invite mode: land the policy and UI topics on main

MrBuondi:
## Why this PR exists

`main` currently has **half** of the group invite mode. #133 landed the data layer, but #134 and #135 merged into their stacked *base branches* instead of into `main`, because those bases were not deleted and GitHub therefore never retargeted them.

Verifiable on `main` right now:

```
lib/models/group_invite_mode.dart          present
lib/util/group_pending_invite_store.dart   present
lib/services/group_invite_promoter.dart    MISSING
lib/screens/invite_requests_screen.dart    MISSING

privacy_settings_screen.dart : 0 occurrences of GroupInviteMode
group_service.dart           : 0 occurrences of GroupPendingInviteStore.hold
```

So `main` has a `chat_app.db` v14 table nothing ever writes and a setting nothing ever reads. This PR lands the remaining two topics and makes the feature whole.

## What is in it

The seven commits already reviewed in #134 and #135, cherry-picked onto `main` in topic order — no merge commits, so the history stays linear and every commit keeps its original author:

| Topic | Commits |
|---|---|
| Service policy (was #134) | hold first-contact invites · promote held invites once the sender becomes known · keep the promoter from applying invites in the strict mode |
| UI (was #135) | mode picker in Privacy Settings · `InviteRequestsScreen` + entry points · discard held invites on switch to strict · refresh the pending-invite counters |

Nothing new: the resulting tree is byte-identical to a straight merge of `invite-mode/policy` into `main`, which is how it was verified.

## The property that decides this feature — do not regress it

This is **not** "reopen the profile fetch". M2 stays closed in **both** modes:

- `_resolveSenderIdentity` remains **DB-only**. No Tor fetch for an unknown sender, on any group control path.
- The held envelope stays encrypted and opaque until the user adds the contact, so the fetch starts from a **user gesture**, never from unauthenticated traffic.
- The two `POLICY:` tests in `test/security/group_profile_fetch_gate_test.dart` are unchanged from #128 — assertions intact. The positive one is load-bearing: without it the negative would pass even if invites were broken outright.

Anyone touching `group_service.dart:500-527` has to preserve this.

## Verification

- `flutter analyze`: 0 issues.
- `flutter test`: **961 passed + 1 skip** (the skip is the L3 harness, gated by `PRYSM_E2E`).
- Tree parity against the merge-based integration checked with `git diff --exit-code`.

## After merging

Delete `invite-mode/data`, `invite-mode/policy` and `invite-mode/ui`. Leaving them is what caused the mis-targeting in the first place.

## Summary by CodeRabbit

* **New Features**
  * Added configurable group invite modes, including contacts-only and pending request handling.
  * Added an Invite Requests screen to review, accept, or discard invitations from unknown senders.
  * Added pending invite counts and navigation entries on the Home and Settings screens.
  * Automatically processes held invites when invite settings allow it.
* **Bug Fixes**
  * Improved handling of unknown senders and invite storage failures without disrupting normal messaging.
  * Preserved existing contact names when adding contacts with empty display names.

### DIFF
--- a/lib/screens/home/home_screen.dart
+++ b/lib/screens/home/home_screen.dart
@@ -46,6 +46,7 @@ import 'package:prysm/screens/group_chat.dart';
 import 'package:prysm/models/conversation.dart';
 import 'package:prysm/models/conversation_preferences.dart';
 import 'package:prysm/models/group.dart';
+import 'package:prysm/models/group_invite_mode.dart';
 import 'package:prysm/services/conversation_preferences_service.dart';
 import 'package:prysm/models/detached_chat_launch.dart';
 import 'package:prysm/models/share_target.dart';
@@ -409,8 +410,11 @@ class _HomeScreenState extends State<HomeScreen> with WidgetsBindingObserver {
     // Invites held while their sender was unknown apply themselves once the
     // contact exists — including when the user added them from somewhere
     // else entirely. Skipped in decoy mode: a panic session must not touch
-    // real group state.
-    if (!widget.decoyMode) {
+    // real group state. Also skipped in contactsOnly mode, where nothing
+    // may be stored or applied: the promoter itself refuses there too, but
+    // the sweep should not even start.
+    if (!widget.decoyMode &&
+        settings.groupInviteMode == GroupInviteMode.holdAsRequest) {
       unawaited(
         GroupInvitePromoter(
           userId: widget.onionAddress,
--- a/lib/screens/home/home_screen.dart
+++ b/lib/screens/home/home_screen.dart
@@ -22,6 +22,7 @@ import 'package:prysm/app/app_composition.dart';
 import 'package:prysm/app/conversation_list_repository.dart';
 import 'package:prysm/app/tor_connection_controller.dart';
 import 'package:prysm/screens/settings_screen.dart';
+import 'package:prysm/screens/invite_requests_screen.dart';
 import 'package:prysm/services/battery_saver_service.dart';
 import 'package:prysm/services/block_service.dart';
 import 'package:prysm/services/active_conversation_tracker.dart';
@@ -73,6 +74,7 @@ import 'package:prysm/ui/core/prysm_pressable.dart';
 import 'package:prysm/util/notification_service.dart';
 import 'package:prysm/util/conversation_refresh_notifier.dart';
 import 'package:prysm/util/group_membership_notifier.dart';
+import 'package:prysm/util/group_pending_invite_store.dart';
 import 'package:prysm/screens/widgets/add_contact_dialog.dart';
 import 'package:prysm/screens/widgets/qr_scanner_screen.dart';
 import 'package:prysm/screens/widgets/prysm_id_qr.dart';
@@ -156,6 +158,7 @@ class _HomeScreenState extends State<HomeScreen> with WidgetsBindingObserver {
   bool _loadUsersQueued = false;
   bool _loadUsersQueuedLight = false;
   bool _sharePickerOpen = false;
+  int _pendingInviteCount = 0;
 
   int get _archivedCount => conversations
       .where((c) => _conversationPrefs[c.id]?.isArchived ?? false)
@@ -180,6 +183,7 @@ class _HomeScreenState extends State<HomeScreen> with WidgetsBindingObserver {
       return 0;
     }
     var count = 0;
+    if (_pendingInviteCount > 0) count++;
     if (_archivedCount > 0) count++;
     if (_blockedCount > 0) count++;
     return count;
@@ -877,6 +881,10 @@ class _HomeScreenState extends State<HomeScreen> with WidgetsBindingObserver {
         _selfChatLastPreview = deferred[3] as String?;
       }
 
+      final pendingInvites = await GroupPendingInviteStore.count();
+      if (!mounted) return;
+      setState(() => _pendingInviteCount = pendingInvites);
+
       _applyLoadedUsers(
         userMaps: userMaps,
         timestamps: timestamps,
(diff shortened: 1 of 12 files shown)


MrBuondi:
All six verified against the code before touching anything. **All six real, all six fixed** — no false positives. One of them is worse than filed, and one suggested remedy is not applicable; details below.

## CR6 — `promote` does not check the mode

Correct, and this is the one that matters. `promoteResolvable` gated the sweep, `promote` did not, and `promote` is what the requests screen calls. The screen holds whatever rows its last load returned, so it survives a mode change made in another route. Fixed in `70a15f9`, with the check **before** `take()` — otherwise the call that is not allowed to apply the invite would still destroy the row on its way to refusing.

The gate now sits on the effect rather than on one caller, which is where this codebase already learned to put it (`group_service.dart`: gate the fetch, not the message). `promoteResolvable` keeps its own check as a fast path that skips the store read, not as the guarantee.

Test: `contactsOnly: a direct promote applies nothing and keeps the row`. Proven mordant — with the gate removed it fails.

## CR2 — `displayName: ''` — real, but worse than filed, and the suggested fix cannot work

The finding says the name "remains empty" if enrichment supplies no username. It is stronger than that: accepting an invite **destroys a name that was already there**.

`insertOrUpdateUser` is `ConflictAlgorithm.replace`, i.e. a whole-row write, and `addContact` hardcodes `name: ''` (it always did — `displayName` only ever feeds `customName`). The row already exists by then, because `ensureUserExist` runs on every inbound handler and seeds `Unknown - xxxxxx`. Proven with a throwaway probe against the real `DBHelper`:

```
PROBE before accept: name=Unknown - abcdef
PROBE after  accept: name= customName=null
```

The proposed remedy — pass `_shortOnion(senderId)` "without storing it as `customName`" — is not reachable through this API: `addContact` maps `displayName` to `customName` and nothing else (`contact_add_service.dart:120`). Passing it would pin a truncated onion as the user's **nickname**, which is exactly the field #132 added for user-chosen names and which permanently outranks `name`. So enrichment could never show through.

Fixed at the source instead, in `03ca57d`: `addContact` preserves a non-empty existing `name` rather than blanking it. The invite flow keeps passing `displayName: ''`, because a nickname is the user's to choose, not this screen's to invent. Test added to the characterization suite; proven mordant.

Note, not changed: `avatarUrl` and `avatarBase64` are clobbered by the same whole-row write. Pre-existing, not reachable through this PR, and enrichment restores the avatar — flagging it rather than widening the diff.

## CR5 — mode changes are reentrant

Correct. The guard at the top compares against `_groupInviteMode`, which the handler has already updated, so a second selection passes it while the first is still suspended on `setGroupInviteMode`. The first then resumes with its own captured `value == contactsOnly` and wipes the store, while the committed mode is the one that allows holding.

Fixed in `6299b8e` on both axes you named: a reentrancy flag drops the overlapping call, the radio rows are disabled while a change is in flight, and the clear decision now reads the **committed** `settings.groupInviteMode` instead of trusting the captured parameter.

## CR3, CR4, CR1 — the requests screen

All three correct, fixed together in `d0f4bf9`:

- **`_load` failure** left `_loading` true forever and let the exception escape the unawaited future in `initState`. Now caught, with a visible error state instead of a permanent spinner.
- **`_accept` discarded `promote`'s bool and had `try`/`finally` with no `catch`.** A throw skipped `_load()` entirely and left the list stale. Now every outcome is reported: contact unreachable, or contact added but the invite could not be applied — the second one matters because `promote` consumes the row either way, so without a message the request just vanishes with no group to show for it. `_load()` runs on every path.
- **`_busySenderId` is a single slot**, so a second tap stole it, dropped the first row's spinner, and raced two accept flows against the same store. Every row is now locked while any accept is in progress (`PrysmTextButton.onPressed: null` already renders and behaves as disabled), plus a re-entry guard in `_accept` itself.

## On tests for the three screen fixes

There are none, deliberately, and not out of laziness. `test/group_invite_mode_screens_test.dart` documents at the top why the requests screen is excluded: its content arrives from a real sqflite read, and under `testWidgets`' fake-async zone that future never completes, so an assertion **hangs instead of failing**. I tried anyway for CR5, with `sqflite_common_ffi` and `setDatabaseForTest` — the run hung and was killed at 180 s. A test that hangs is worse than no test, so it was removed rather than committed.

The two findings with a reachable seam (CR6, CR2) do have tests, and both were proven mordant against the pre-fix code.

## Gate

`flutter analyze` 0 issues · `flutter test` **963 passed + 1 skip** (was 961; +2 from the new tests).

MrBuondi:
Correct, and it caught a real gap in my own fix — I preserved `name` and left `customName` being nulled. Worse, in my previous reply I explicitly waved off the neighbouring columns as "enrichment restores the avatar", which does not apply to a nickname at all: `_enrichFromProfile` writes only `name` and `avatarBase64`, so **nothing** ever restores `customName`. Re-adding an existing contact through the invite flow would have wiped a user-chosen nickname permanently.

Fixed in `a17ea47`, and widened to the whole row rather than to the one column you named, because the defect is the write shape, not the field: `insertOrUpdateUser` is `INSERT OR REPLACE`, so every column not carried over is destroyed. `addContact` now reads the row and merges, overwriting only the identity — which is the one thing it is actually there to update.

This is not a new pattern: `PeerIdentityResolver` already does exactly this for the same operation (`peer_identity_resolver.dart:113-121`), and `addContact` was the odd one out.

```dart
name:         (existing?['name'] as String?) ?? '',
avatarUrl:    (existing?['avatarUrl'] as String?) ?? '',
avatarBase64: existing?['avatarBase64'] as String?,
customName:   displayName.isNotEmpty
                  ? displayName
                  : existing?['customName'] as String?,
identityJson: identityJson,   // the only intentional overwrite
```

Two tests, both proven mordant against the pre-fix code:

- `an empty displayName keeps the existing nickname and avatar instead of nulling them` — seeds `name: 'Alice B.'`, `customName: 'Bob'`, `avatarBase64: 'YWJj'`, re-adds with `displayName: ''`, asserts all three survive **and** that `identityJson` was still replaced.
- `a non-empty displayName still replaces the nickname` — the guard against over-correcting: an explicit name from the caller must still win, or the merge would make nicknames unchangeable.

Gate: `flutter analyze` 0 issues · `flutter test` **965 passed + 1 skip** (was 963).

PR #138, opened 2026-08-07
feat(search): implement full-text search for messages

xmreur:
- Added MessageSearchDao for managing FTS5-backed local message search.
- Introduced MessageSearchBackfillService to index historical messages for search.
- Updated MessageCrudDao to integrate search functionality, including message removal.
- Enhanced MessagesDb with global and conversation-specific search methods.
- Implemented UI components for search in chat screens, including ChatSearchBar.
- Updated database schema to support message search indexing.
- Added MessageSearchHit model for search results representation.

## Summary by CodeRabbit

* **New Features**
  * Added global message search with snippets, grouped conversations, loading states, and navigation to matching messages.
  * Added in-conversation search with match counts, previous/next controls, highlighted matches, and direct navigation.
  * Search covers text and file names across direct, group, and self chats.
  * Added automatic indexing for new and edited messages, plus background indexing of existing messages.

* **Bug Fixes**
  * Deleted and view-once messages are removed from search results.
  * Blocked conversations no longer show message previews or unread badges.

### DIFF
--- a/lib/app/app_composition.dart
+++ b/lib/app/app_composition.dart
@@ -3,6 +3,7 @@ import 'package:prysm/server/PrysmServer.dart';
 import 'package:prysm/services/block_service.dart';
 import 'package:prysm/services/call/call_manager.dart';
 import 'package:prysm/services/file_transfer_handler.dart';
+import 'package:prysm/services/message_search_backfill_service.dart';
 import 'package:prysm/services/read_receipt_service.dart';
 import 'package:prysm/services/sync_coordinator.dart';
 import 'package:prysm/services/wake_hint_service.dart';
@@ -106,5 +107,16 @@ class AppComposition {
   /// needs to know whether Tor was intentionally stopped.
   static void wireTorRuntimeGate(bool Function() isTorStopped) {
     TorRuntimeGate.isTorStopped = isTorStopped;
+  }
+
+  /// Starts background indexing of historical messages for FTS search.
+  static void startSearchBackfill({
+    required KeyManager keyManager,
+    required String userId,
+  }) {
+    MessageSearchBackfillService(
+      keyManager: keyManager,
+      userId: userId,
+    ).startIfNeeded();
   }
 }
--- a/lib/app/app_composition.dart
+++ b/lib/app/app_composition.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:prysm/crypto/peer_proof.dart';
 import 'package:prysm/server/PrysmServer.dart';
 import 'package:prysm/services/block_service.dart';
@@ -10,6 +12,7 @@ import 'package:prysm/services/wake_hint_service.dart';
 import 'package:prysm/transport/transport_provider.dart';
 import 'package:prysm/util/key_manager.dart';
 import 'package:prysm/util/local_onion_address.dart';
+import 'package:prysm/util/logging.dart';
 import 'package:prysm/util/tor_runtime_gate.dart';
 import 'package:prysm/util/tor_service.dart';
 
@@ -114,9 +117,12 @@ class AppComposition {
     required KeyManager keyManager,
     required String userId,
   }) {
-    MessageSearchBackfillService(
+    final backfill = MessageSearchBackfillService(
       keyManager: keyManager,
       userId: userId,
     ).startIfNeeded();
+    unawaited(backfill.catchError((Object e, StackTrace stack) {
+      Logging.error('Search backfill failed: $e\n$stack', 'SearchBackfill');
+    }));
   }
 }
--- a/lib/database/conversation_queries_dao.dart
+++ b/lib/database/conversation_queries_dao.dart
@@ -1,11 +1,15 @@
 import 'package:prysm/database/message_query_filters.dart';
+import 'package:prysm/database/message_search_dao.dart';
 import 'package:prysm/database/messages_database.dart';
 import 'package:sqflite/sqflite.dart';
 
 /// Direct/group conversation message queries (batched + paginated reads) and
 /// their bulk deletes.
 class ConversationQueriesDao {
-  const ConversationQueriesDao();
+  const ConversationQueriesDao({MessageSearchDao? searchDao})
+      : _searchDao = searchDao ?? const MessageSearchDao();
+
+  final MessageSearchDao _searchDao;
 
   Future<Database> get _database => MessagesDatabase.database;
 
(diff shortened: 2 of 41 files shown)


MrBuondi:
Full review of this PR, with the fixes for everything blocking in #139 (based on
`feat/message-search`, so merging it here carries them into this PR).

Reviewed at `522e38b` merged with current `main` — the branch forked at `4f80b5e`, before #133,
#136 and #137 landed. The merge is clean. `flutter analyze` 0 issues, `flutter test` 987 + 1 skip
on that integration.

The feature works: I drove it on the real app under `tool/live/prysmlab` — global search, the
in-conversation bar with its match counter and prev/next, persistence across a restart, and six
hostile FTS5 metacharacter queries with no crash and no syntax error. What follows is what sits
underneath that.

## Blocking — the index is never deleted

The PR adds a **plaintext searchable copy of every message body** and then misses three of the
delete paths, so content the user erased stays readable in the sidebar search, across restarts.

| | Finding | Where |
|---|---|---|
| B1 | `deleteMessageById` never de-indexes — the branch every **received** message's delete takes, since delete-for-everyone requires being the sender | `message_crud_dao.dart:296-309` ← `message_actions_service.dart:91,100` |
| B2 | Leaving or deleting a group leaves every group body indexed; likewise the rejoin purge and "clear chat" on a direct conversation | `conversation_queries_dao.dart:100-113, 154-173` ← `group_service.dart:456,696` |
| B3 | A view-once is searchable **before** it is viewed: the gate is `viewOnce && viewed`. The sender's own copy can never be opened, so it is never de-indexed at all | `message_search_index_service.dart:171` |

B1 proven on the running app, against the real SQLCipher database:

```
PROBE_ROWS_BEFORE=1
PROBE_ROWS_AFTER=0
PROBE_FTS_AFTER_DELETE=1 body=confidential axolotl dossier from a peer

$ prysmlab count "confidential axolotl dossier from a peer"
   1        # still painted in the sidebar, after the message row is gone
```

It survived an app restart. At DAO level the same probe is red for `deleteMessageById`,
`deleteMessagesForGroup`, `deleteGroupMessagesBefore` and end-to-end through
`MessageActionsService.deleteMessage`, while the `softDeleteMessage` / `hardDeleteMessage`
controls pass — same seed, same search, only the delete API differs.

## Also blocking

**The backfill dies on any message over 512 KiB.** `_resolveMessageBlobs` mutates rows returned
by `db.rawQuery`, which sqflite returns read-only:

```
PROBE rawQuery mutate -> THROWS: Unsupported operation: read-only
```

The whole run throws before indexing anything, the cursor stays at 0, `complete` is never set,
and it crashes again on every launch. A user with one large attachment never gets a historical
index at all. On the live path the same marker is handed straight to the decrypt, which throws
and is swallowed, so those messages are not indexed there either.
`message_search_backfill_service.dart:314-324`, `inbound_message_router.dart:677-683`.

## Important

- **The direct-path FTS removal is unscoped and remotely triggerable.** `_deleteRow` falls back
  to `WHERE messageId = ?`, and group wire ids travel in cleartext to every member
  (`group_chat_service.dart:584-601`) while `_isValidMessageData` only checks `is String`
  (`inbound_message_router.dart:859-867`). A member can reuse an observed id in a direct message
  and wipe the victim's group search rows the moment they delete it — or merely open it, if it is
  view-once. Verified: the group hit went `1 → 0` while the group `messages` row still existed.
- **Five index call sites are not best-effort**, contradicting the contract the service documents
  at `message_search_index_service.dart:48-50`: `self_chat_service.dart:38-45, 78-85` and
  `message_modify_service.dart:155-162, 211-218, 434-442`. With the FTS table missing all five
  throw **after** the row is persisted; four reach the UI uncaught, the fifth turns an applied
  inbound edit into a 500.
- **Message bodies stopped honouring the system font size.** `linked_message_text.dart:84` now
  always returns `RichText` (`TextScaler.noScaling`) where the pre-PR code returned `Text` for
  URL-free bodies. Measured in a widget test: 14.0 under a 2.0 scaler, against 28.0 for a bare
  `Text`.
- **The backfill can mark itself complete with rows never indexed.** The per-row failure counter
  is persisted in SharedPreferences and never reset on success, so five failures spread over five
  launches — possibly five different transient causes — burn a row permanently.

All of the above are fixed in #139, each with a test that fails on the pre-fix code (13 of 13
bite), plus live verification on the real app.

## Minor — not in #139, listed so they are not lost

- `buildSnippet` can exceed `maxLen` by 2: the ellipses are added before the clamp.
- `İ` (U+0130) lowercases to two code units and mis-centres the snippet window; `toLowerCase`
  does no diacritic folding while FTS5 uses `remove_diacritics 2`, so a hit can render with no
  visible match. Clamped, so cosmetic.
- `ChatSearchBar` does not wrap `MessagesDb.searchMessagesInConversation` in try/catch, and it
  keeps the previous query's results live during the debounce, auto-selecting the last match on
  every completion.
- Blocking a contact does not remove hits already on screen until the next keystroke.
- **No on-disk upgrade test for `messages.db` v13→v14.** `test/message_schema_migrations_test.dart`
  is in-memory; the on-disk `chat_db_v13_v14_upgrade_test.dart` covers the other database. This one
  is a repo convention, not a preference.
- The backfill decrypts 200 rows per batch on the UI isolate with a 16 ms yield, started from
  `HomeScreen.initState`.
- No size cap on an indexed body, where `GroupPendingInviteStore` sets the precedent at 64 KiB.
  Note the trap it already documents: measuring with `utf8.encode` on untrusted input is itself
  the DoS.
- `MessageSearchIndexService.removeMessage` has no callers.
- Every `upsert` scans the whole FTS table — all columns are `UNINDEXED`. Measured 0.59 / 1.0 /
  1.9 ms at 5k / 10k / 20k rows, linear.
- Pre-existing, outside this PR but adjacent: `insertInboundMessage` can be clobbered by a
  sender-chosen id, and `deleteMessageById` nulls `replyTo` by comparing a storage id against
  wire ids, which is a no-op for group rows.

## On the CodeRabbit rounds

I re-verified all 37 findings against `522e38b`. 35 are real and already resolved here,
including all five delivery-critical-path sites, the non-reentrant mutex nesting, the empty-body
upsert retention and all eight nitpicks. Two are wrong: the Critical "convert the clamp result to
int" (`int.clamp(int, int)` is statically `int`; `flutter analyze` is clean and a minimal repro
confirms it) and the `localId`/`receiverId` item (pre-existing, and the new `localUserId` never
participates in the FTS row identity). One is only partially fixed: the conversation-scoped index
identity landed for group-vs-group with tests, but the direct path still deletes unscoped.

Worth saying plainly: all 37 concern how the index is **written**. None concerns whether it is
**deleted** — which is where the feature breaks.

PR #140, opened 2026-08-09
perf(ratchet): read the peer scheme from cache on the send path

MrBuondi:
## Problem

`RatchetService.encryptBytes` runs inside the per-peer ratchet lock. When no session exists yet
(the first message to a peer) it awaited `TransportProvider.getProfileOrFallback` — `timeout: 20s`,
`maxAttempts: 3`, each failed attempt able to force a global `SIGNAL NEWNYM` — purely to read an
optional `ratchetScheme` hint. On any error the result is discarded, so the cost was mandatory and
the value was not.

Measured latency passthrough, first send vs stubbed fetch delay:

| fetch delay | 0 ms | 250 ms | 1000 ms | 3000 ms |
|---|---|---|---|---|
| send latency | 17 ms | 255 ms | 1006 ms | 3005 ms |

Introduced in v0.6.4; the code is absent from v0.6.3.

## Change

- Persist the peer's scheme on the `users` row (`chat_app.db` v14 -> v15), warmed by the profile
  fetches the app already performs (contact add, peer identity resolution, chat profile refresh,
  inbound sender refresh).
- The send path reads that cache. A cold cache makes a **single** bounded attempt (3s, no circuit
  refresh) and falls back to the v2 initializer, re-warming in the background.
- The bound is enforced on the await itself, not only passed to the transport, because the device
  log shows Tor futures that never complete.

Ratchet v3 is a real forward-secrecy improvement over v2, so the cold path still attempts the
fetch rather than downgrading silently. The downgrade fallback itself predates this change.

## Verification

- Warm cache **12 ms**, still negotiates v3. Cold cache against a profile that never answers
  **3031 ms**, bounded (previously 3 x 20s plus a forced circuit reset).
- Migration exercised on a real SQLCipher database in the running Linux app: `user_version = 15`,
  `users.ratchetScheme` present.

## Summary by CodeRabbit

* **New Features**
  * Added support for discovering and caching peers’ ratchet schemes.
  * Improved session setup with cached negotiation, bounded profile refreshes, and fallback compatibility.
  * Preserved advertised ratchet schemes during contact and profile updates.

* **Bug Fixes**
  * Prevented repeated or unbounded profile requests when scheme discovery fails.
  * Retained existing cached schemes when updated profiles omit this information.

* **Improvements**
  * Added a database upgrade to store cached ratchet scheme information.
  * Expanded coverage for negotiation, fallback, caching, and database migration scenarios.

### DIFF
--- a/lib/crypto/constants.dart
+++ b/lib/crypto/constants.dart
@@ -60,9 +60,11 @@ class CryptoConstants {
     return ratchetSchemeRank(a) >= ratchetSchemeRank(b) ? a : b;
   }
 
-  /// Validates a profile or wire ratchet scheme string.
-  static String? parseRatchetScheme(String? raw) {
-    if (raw == null) return null;
+  /// Validates a profile or wire ratchet scheme string. Peer-controlled
+  /// profile JSON may carry a non-string value (a number or object); those
+  /// are ignored (null), never thrown on.
+  static String? parseRatchetScheme(Object? raw) {
+    if (raw is! String) return null;
     final trimmed = raw.trim();
     switch (trimmed) {
       case schemeRatchet1:
--- a/lib/crypto/ratchet/ratchet_service.dart
+++ b/lib/crypto/ratchet/ratchet_service.dart
@@ -1,3 +1,4 @@
+import 'dart:async';
 import 'dart:convert';
 
 import 'package:cryptography/cryptography.dart';
@@ -9,7 +10,9 @@ import 'package:prysm/crypto/ratchet/prekey_bundle.dart';
 import 'package:prysm/crypto/ratchet/ratchet_session.dart';
 import 'package:prysm/crypto/ratchet/session_store.dart';
 import 'package:prysm/crypto/wire.dart';
+import 'package:prysm/transport/transport_preference.dart';
 import 'package:prysm/transport/transport_provider.dart';
+import 'package:prysm/util/db_helper.dart';
 
 /// High-level 1:1 ratchet encrypt/decrypt with SQLite session persistence.
 class RatchetService {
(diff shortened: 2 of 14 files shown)


MrBuondi:
All four actionable findings plus the nitpick were verified against the code and fixed in `d78b587`. Full suite green, `flutter analyze` clean.

**Share an in-flight scheme fetch after the await timeout** — correct, and the sharpest finding here. The foreground fetch was untracked: `_inFlightSchemeWarms` was only added to inside `_warmPeerRatchetScheme`, so a cold bootstrap started fetch #1, the 3s timeout fired, and the fallback started fetch #2 while #1 was still running. Repeated cold bootstraps kept adding untracked foreground fetches. Replaced the `Set<String>` with a `Map<String, Future<String?>>`: one shared future per peer, persistence chained onto it so a timed-out caller still warms the cache, and the map entry cleared on settle with an identity check so a newer fetch is not evicted by an older one completing. Regression test uses a `Completer`-backed fetcher and counts invocations across three cold bootstraps: **pre-fix 4, post-fix 1**.

**Ignore non-string advertised scheme values** — correct, and this JSON is peer-controlled. Fixed at the root rather than at the four call sites: `CryptoConstants.parseRatchetScheme` now takes `Object?` and returns null for anything that is not a `String`, and the casts are gone from `peer_identity_resolver.dart`, `chat.dart`, `PrysmServer.dart` and `contact_add_service.dart`. Test: a profile with a numeric `ratchetScheme` is ignored while identity, name and avatar are still applied — pre-fix it threw `type 'int' is not a subtype of type 'String?' in type cast` and dropped the whole profile.

**Close or isolate the in-memory database between tests** — correct. `tearDown` only cleared `DBHelper`'s reference, so `singleInstance: true` handed the next `setUp` the same `:memory:` database with the previous test's `users` rows. The code was working around this by explicitly overwriting the row, which is exactly the kind of workaround that rots; the database is now closed in an async `tearDown` and the defensive overwrite is gone.

**Assert the v15 user-column migration on both legacy paths** — correct, and not hypothetical: a real bug in this exact step was fixed during development, where the `ALTER TABLE users` fired against databases with no `users` table because `PRAGMA table_info` returns an empty list rather than erroring. Both fixtures do create `users`, so the assertion is valid on both paths and is now present.

**Assert the original user values after migration** (nitpick) — correct, applied: `id`, `name` and the null `ratchetScheme` are asserted, so a migration that deleted and re-inserted a row no longer passes.

PR #141, opened 2026-08-09
fix(tor): stop offline peers from degrading healthy ones

MrBuondi:
## Problem

`TorDelivery.withTorRetry` called `_maybeRefreshCircuit(circuitError: true, force: i == 0)`, and
`isCircuitError` returned true for SOCKS `hostUnreachable`. A permanently offline peer fails
immediately with `hostUnreachable`, so **every** delivery attempt to a dead contact forced a global
`SIGNAL NEWNYM`, bypassing the rate limit and tearing down the circuits of reachable peers.

From the device log: 80 `hostUnreachable` failures, 116 `WebSocket not connected`, 63
`WebSocket disconnected`; the one reachable peer repeatedly loses its WebSocket and falls back to
HTTP. At startup six peers are wake-hinted and five are unreachable, several after a 15s timeout.

## Change

- A peer being down is not a circuit fault: drop `hostUnreachable` / `networkUnreachable` from the
  circuit-error set. They stay in `isRetryableError`, so retry behaviour is unchanged; only the
  NEWNYM classification differs. `ttlExpired` and SOCKS server failures remain circuit errors.
- Remove the `force` bypass so the rate limit always applies: a routine per-peer delivery failure
  can no longer force a global circuit reset.
- Skip wake hints for peers that are repeatedly unreachable, not merely once.
- A backoff fast-fail no longer counts as a fresh connect failure, so a peer in backoff is still
  re-dialled instead of having its window slid forward on every heartbeat.
- A user-initiated send to a peer in backoff fails fast to the queue/HTTP route instead of blocking
  the UI for the full interactive budget.

## Verification

- Tests drive the real maintenance loop rather than the test-only injection hook, so they fail if
  the production path stops reaching the behaviour.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced unnecessary connection attempts to peers that remain unreachable.
  * Wake hints are now suppressed for persistently unreachable peers.
  * Incoming authenticated hints can restore a peer’s connection availability.
  * Improved handling of network failures without triggering unnecessary Tor circuit refreshes.
  * Connection retry behavior now fails fast during temporary backoff periods.

### DIFF
--- a/lib/services/wake_hint_service.dart
+++ b/lib/services/wake_hint_service.dart
@@ -105,6 +105,11 @@ class WakeHintService {
           .take(BatterySaverPolicy.wakeHintMaxPeers)
           .map((e) => e.key)
           .where((id) => id != userId)
+          // Don't wake peers the connection manager is already throttling:
+          // each hint to a dead peer burns a full HTTP timeout.
+          .where(
+            (id) => !TransportProvider.instance.wsManager.isPeerUnreachable(id),
+          )
           .toSet();
     } catch (_) {
       return;
@@ -152,6 +157,9 @@ class WakeHintService {
     _lastReceivedFromPeer[senderId] = DateTime.now();
 
     if (TransportProvider.isConfigured) {
+      // The sender proved reachability with an authenticated hint: lift any
+      // backoff/quarantine so the flush connect is not rate-limited.
+      TransportProvider.instance.wsManager.clearPeerFailureState(senderId);
       unawaited(
         TransportProvider.instance.wsManager
             .ensureConnected(senderId)
--- a/lib/services/wake_hint_service.dart
+++ b/lib/services/wake_hint_service.dart
@@ -102,14 +102,16 @@ class WakeHintService {
       final recent = timestamps.entries.toList()
         ..sort((a, b) => b.value.compareTo(a.value));
       peers = recent
-          .take(BatterySaverPolicy.wakeHintMaxPeers)
           .map((e) => e.key)
           .where((id) => id != userId)
           // Don't wake peers the connection manager is already throttling:
-          // each hint to a dead peer burns a full HTTP timeout.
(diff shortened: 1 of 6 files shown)


MrBuondi:
Both findings were verified and fixed in `6126f96`. Full suite green, `flutter analyze` clean.

**Filter unreachable peers before `take`** — correct, and it matters more than the severity suggests. In the device log that motivated this PR, five of six contacts are dead, so the cap was being spent on peers that would never answer while the one reachable peer went unhinted. Reordered so the limit counts only peers that will actually be hinted. Test seeds 30 peers with the five most recent throttled: **pre-fix 14 hints, post-fix 20**.

**`_quarantineUntil` is written but never read by the gating predicates** — the substance is right and is fixed, but the headline is not accurate and I would rather say so than nod along. `_quarantineUntil` *is* read: `_isQuarantined` at `ws_connection_manager.dart:274`, called from `shouldDialPeer` at `:256`. The real defect is narrower and is the one your bullets describe: the quarantine depended on `_quarantineUntil` and `_nextRetryAfter` staying in sync, and they did not.

Three concrete divergences, all now closed:
- `stop()` (`:157`) did `_nextRetryAfter.clear()` and left `_quarantineUntil` populated.
- `isPeerUnreachable` required at least 60s of remaining window, so wake hints resumed during the final minute of a 10-minute quarantine.
- The success path (`:217-219`) removed `_nextRetryAfter` but not the quarantine, so a peer that had just proved it was alive stayed quarantined.

`_isInBackoff` and `isPeerUnreachable` now both return true while `_isQuarantined` holds, which makes the quarantine authoritative rather than a function of bookkeeping symmetry, and the proven-reachable path clears it. The new tests drive real dial failures through the maintenance loop rather than the `@visibleForTesting` hook, and the existing test that depended on `stop()` clearing only `_nextRetryAfter` was rewritten so it no longer relies on that asymmetry.

PR #142, opened 2026-08-09
perf(search): make the FTS upsert cost independent of history size

MrBuondi:
## Problem

Every metadata column of `message_search_fts` is declared `UNINDEXED`, so there is no b-tree on
`(messageId, conversationId, scope)`. The DELETE half of each upsert therefore full-scanned the
index, as did `exists()`.

| index rows | 5k | 10k | 20k | 40k |
|---|---|---|---|---|
| ms per upsert | 0.62 | 1.08 | 2.05 | 3.96 |

Linear, and paid inline on both the send and the receive path while holding the single global
`MessagesDatabase.mutex` that serialises all `messages.db` work.

## Change

- Add `message_search_rows`, a plain table keyed on `(messageId, conversationId, scope)` carrying
  the FTS rowid, so the row is located by a b-tree probe instead of a scan.
- Schema migration backfills it from the existing index; existing installs keep their index.
- Each mutating pair runs in a single transaction, and the rowid is checked against the intended
  row before deletion. FTS5 reuses rowids, so a stale side row would otherwise delete a different
  message's index entry and leave the intended one searchable.
- The missing-side-row fallback to the legacy predicate is kept, so a delete can never leave an
  orphan index entry behind.

Metadata only: no message body is stored in the side table, and it lives in the same encrypted
database.

## Verification

- **2.05 ms -> 0.39 ms** at 20k rows, and flat with index size instead of linear.
- Exercised on a real SQLCipher database in the running Linux app: sending a message yields one FTS
  row and one side row with the correct body; search finds it; deleting yields zero and zero, and
  the text is no longer searchable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message search index consistency during message and conversation updates or deletions.
  * Prevented stale search data from affecting unrelated messages.
  * Added safer recovery when search index operations encounter errors.

* **Maintenance**
  * Existing databases are upgraded automatically to support the improved search indexing structure.
  * Added coverage for migration, consistency, deletion, scaling, and rollback scenarios.

### DIFF
--- a/lib/database/message_schema_migrations.dart
+++ b/lib/database/message_schema_migrations.dart
@@ -11,7 +11,7 @@ class MessageSchemaMigrations {
   MessageSchemaMigrations._();
 
   /// Current schema version. Bump alongside a new _upgradeToVN step.
-  static const int dbVersion = 14;
+  static const int dbVersion = 15;
 
   static Future<void> onCreate(Database db, int version) async {
     await _createV2(db);
@@ -34,6 +34,7 @@ class MessageSchemaMigrations {
     if (oldVersion < 12) await _upgradeToV12(db);
     if (oldVersion < 13) await _upgradeToV13(db);
     if (oldVersion < 14) await _upgradeToV14(db);
+    if (oldVersion < 15) await _upgradeToV15(db);
   }
 
   static Future<void> migrateOversizedMessagePayloads(Database db) async {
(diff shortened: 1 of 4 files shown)


MrBuondi:
Two of the three nitpicks are applied in `6b21e74`; the index is declined, with reasoning. Full suite green, `flutter analyze` clean.

**Guard the timing ratio against a very small denominator** — applied, and chasing it turned up something worse than the flake you flagged. Measuring both paths on this machine: the rowid probe runs at **0.87–0.95x** and a forced O(N) predicate path at **3.18–3.79x**. The bound was `4x`, so the assertion passed with the regression in place — I confirmed it, forced O(N) passed 9 runs out of 9. The test was not merely flaky, it could not fail. Retuned to `3x`, verified that forced O(N) now fails (`2338.7us > 2127.6us`), and added the denominator floor at 20µs so a tiny baseline cannot turn scheduler noise into a failure. The floor is inert at the current scale (`smallPerOp` ≈ 0.6ms) and is there for the small-index case you described.

**Document the transaction constraint on `removeForConversationUnprotected`** — applied, same doc shape as `removeUnprotected`, and all three call sites were checked to comply.

**Index on `message_search_rows(conversationId, scope, timestamp)`** — declined. Two reasons, both checked against the code:

1. It cannot make the conversation-wide delete proportional. The same transaction deletes from `message_search_fts` with the same predicate, and that table's metadata columns are all `UNINDEXED` (`message_schema_migrations.dart:288-295`), so that half is a full scan regardless. The index would remove the cheaper of two scans and leave the expensive one.
2. It would be maintained on every insert into and delete from `message_search_rows` (`message_search_dao.dart:61`, `:147-151`) — that is, on every message sent and received. This PR exists to take per-message cost *off* the hot path; adding write amplification there to speed up a rare user-initiated operation (clear chat, leave group) inverts the trade it is making.

If conversation-wide deletes become a measured problem, the fix is on the FTS side, not the side table.

PR #143, opened 2026-08-09
fix(modify): stop dropping a peer's delete silently

MrBuondi:
## Problem

Reported as "my friend cannot delete messages". The delete UI is fine; the propagated delete was
being lost without a trace, on four independent counts:

- `MessageModifyService.applyInbound` swallowed every decrypt failure and returned silently.
- `WsFrameRouter` acked `message_modify` **before** processing it, so the sender's transport
  reported success regardless of the outcome.
- The HTTP fallback never inspected the status code — `grep statusCode` over the HTTP transport
  matched nothing — so a 4xx rejection was read as a body and reported as a successful send. This
  is the path a delete takes whenever the peer's WebSocket is down.
- `deleteMessage` discarded the transport result and returned a hardcoded `true`.

Net effect: the deleter saw a tombstone, the recipient kept the message, and nothing anywhere
reported a failure.

## Change

- Categorise and log inbound modify failures, distinguishing "could not authenticate" from
  "unknown target" from "ownership rejected".
- Verify-then-ack for `message_modify` only, so a rejection reaches the sender and the existing
  queue/retry machinery engages. Ordinary `message` frames keep the optimistic ack.
- Surface non-2xx from the HTTP transport so the fallback path behaves like the WebSocket path.
- Propagate the real propagation result out of the direct and group modify paths into the failure
  state the screens already render.
- Ack `unknownTarget` as the benign no-op its own enum documents, instead of returning a 4xx that
  queued a pending row and re-sent it on every sync cycle forever — the case where the peer
  legitimately no longer holds the message.

## Security

The unsigned-envelope gate is **unchanged**. A delete that cannot be authenticated is still
rejected; it is simply no longer rejected in silence. `allowLegacyUnsignedDhAead` is untouched.

A peer on <= v0.6.3 emits legacy unsigned envelopes, so its deletes remain refused. That is the
intended outcome; the change makes the refusal visible instead of invisible.

## Verification

`flutter analyze` clean; full suite green on this branch. Each behaviour has a test that fails on
the pre-fix code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message deletion failure handling by preserving local tombstones, marking failed deletions, and notifying users.
  * Improved message delivery status reporting, including queued and failed operations.
  * Added clearer handling for invalid, undecryptable, unauthorized, or missing message modifications.
  * Prevented unsuccessful network responses from being reported as successful deliveries.
  * Improved real-time message modification acknowledgments and error reporting.

* **Tests**
  * Added comprehensive coverage for deletion failures, message modifications, transport errors, and acknowledgment behavior.

### DIFF
--- a/lib/screens/chat.dart
+++ b/lib/screens/chat.dart
@@ -1488,11 +1488,26 @@ class _ChatScreenState extends State<ChatScreen> {
     setState(() {
       if (outcome == MessageDeleteOutcome.markedDeletedForEveryone) {
         _messages.updateMessage(message, markMessageDeleted(message));
+      } else if (outcome ==
+          MessageDeleteOutcome.markedDeletedForEveryoneFailed) {
+        // The tombstone was applied locally, but the peer was not notified:
+        // surface the send-failure state (metadata['failed'] -> 'Failed').
+        _messages.updateMessage(
+          message,
+          markMessageDeleted(
+            message.copyWith(
+              metadata: {...?message.metadata, 'failed': true},
+            ),
+          ),
+        );
       } else {
         _messages.removeMessage(message);
       }
       selectedMessageIds.remove(message.id);
     });
+    if (outcome == MessageDeleteOutcome.markedDeletedForEveryoneFailed) {
+      showPrysmToast(context, 'Could not delete for everyone');
+    }
   }
 
   void _resendMessage(Message message) {
@@ -1542,6 +1557,16 @@ class _ChatScreenState extends State<ChatScreen> {
         setState(() {
           if (outcome == MessageDeleteOutcome.markedDeletedForEveryone) {
             _messages.updateMessage(message, markMessageDeleted(message));
+          } else if (outcome ==
+              MessageDeleteOutcome.markedDeletedForEveryoneFailed) {
+            _messages.updateMessage(
+              message,
+              markMessageDeleted(
+                message.copyWith(
+                  metadata: {...?message.metadata, 'failed': true},
+                ),
+              ),
+            );
           } else {
             _messages.removeMessage(message);
           }
--- a/lib/screens/chat.dart
+++ b/lib/screens/chat.dart
@@ -1478,13 +1478,18 @@ class _ChatScreenState extends State<ChatScreen> {
     );
   }
 
-  Future<void> _deleteMessage(Message message) async {
+  /// Deletes [message] and returns the outcome, so the bulk path can report a
+  /// single aggregated failure instead of one toast per selected message.
+  Future<MessageDeleteOutcome> _deleteMessage(
+    Message message, {
+    bool showFailureToast = true,
+  }) async {
     final outcome = await _actionsService.deleteMessage(
       message,
       localUserId: widget.userId,
     );
     _messageCache.remove(message.id);
-    if (!mounted) return;
+    if (!mounted) return outcome;
     setState(() {
       if (outcome == MessageDeleteOutcome.markedDeletedForEveryone) {
         _messages.updateMessage(message, markMessageDeleted(message));
@@ -1505,9 +1510,11 @@ class _ChatScreenState extends State<ChatScreen> {
       }
       selectedMessageIds.remove(message.id);
     });
-    if (outcome == MessageDeleteOutcome.markedDeletedForEveryoneFailed) {
+    if (showFailureToast &&
+        outcome == MessageDeleteOutcome.markedDeletedForEveryoneFailed) {
       showPrysmToast(context, 'Could not delete for everyone');
     }
+    return outcome;
   }
 
   void _resendMessage(Message message) {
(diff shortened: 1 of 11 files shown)


MrBuondi:
@coderabbitai review

xmreur:
@coderabbitai review

MrBuondi:
Thanks — all three inline findings were real. Two are fixed as filed; one is fixed with a stronger remedy than the one proposed, because the analysis under it holds only for a benign response body. Four of the five nitpicks are in; one is declined with evidence.

Gate on the branch: `flutter analyze` 0 issues, `flutter test` **1020 passed + 1 skip** (was 1014 + 1).

## Inline findings

### 1. Bulk delete showed one toast per failed message — fixed, and extended to the 1:1 screen (`caa8d19`)

Confirmed: `_deleteSelectedMessages` calls `_deleteMessage` per id and each failure raised its own toast. `_deleteMessage` now returns the outcome and takes `showFailureToast`, and the bulk caller aggregates into a single toast.

The same audit found the 1:1 path in a worse state: `ChatScreen.deleteSelectedMessages` was a verbatim copy of `_deleteMessage`'s outcome block that showed **no** feedback at all on failure — a silent bulk failure in the PR whose whole point is that a failed delete must be visible. It now delegates to `_deleteMessage` (one copy of the block instead of two) and aggregates the same way.

### 2. Malformed payload — classified explicitly (`677014e`), but the retry-loop premise does not hold

The defect is real: `MessageModifyPayload.decode` throws on a payload an authenticated peer can still send, the throw reaches the catch-all in `_handleMessageModify`, and we answer **500** — we blame our own processing for input we can never accept.

The stated consequence, however, does not survive a check. Returning `decryptFailed` (400) does **not** stop the retry the finding describes:

* `SideChannelTransport.sendDirectAndQueue` enqueues on *any* failed send (`side_channel_transport.dart:79-88`);
* the flushers remove a row **only** on success (`:179-182`, `:209-212`, `:246-249`).

So a 400 is queued and re-flushed exactly like a 500. The only status that avoids the outbox is a 2xx — which is precisely why `unknownTarget` is a 2xx noop, as the test at `test/message_modify_service_test.dart:515-517` already states.

A malformed payload is not in the same class as `unknownTarget`: `unknownTarget` is benign (there is nothing left to delete), while a malformed payload means the sender's delete **was not applied and the message is still here**. Answering 2xx would make it silently look delivered — the exact failure mode this PR closes. It therefore stays a 4xx, in the same class as `decryptFailed`, and the unbounded side-channel outbox retry is pre-existing behaviour shared with `decryptFailed` and with reactions/receipts/timers; bounding it is a separate change, not this PR's.

What the fix buys is what it should: a dedicated `InboundModifyOutcome.malformedPayload`, a log line that names the real cause, and a 4xx that blames the input instead of a 500 that blames us. Two tests, both proven to fail against the pre-fix code (the outcome test threw out of `applyInbound`; the router test observed `500` where it now observes `400`).

### 3. Peer-controlled body — removed from the error, not truncated (`dc033f7`)

Accepted, and the exposure is larger than filed. The probe in the analysis concluded "`StateError` is not retryable" from a *benign* body; the body is chosen by the peer. `TorDelivery` classifies by substring on `error.toString()` (`tor_delivery.dart:29-58`), so a peer answering `4xx` with `hostUnreachable` in its body makes **both** `isRetryableError` and `isCircuitError` true: three attempts plus a forced global `SIGNAL NEWNYM` on the first retry (`withTorRetry:116-118`) — remotely triggerable circuit churn, i.e. the amplifier #141 was written to remove, reachable by anyone who can answer an HTTP request.

Truncating does not close it: `hostUnreachable` is 15 characters. The body is therefore not carried at all. `postJson` now throws `PeerHttpStatusException(statusCode)`, whose `toString()` is ours; the body is still drained (socket release) and discarded. The class deliberately does not extend `HttpException`, which `isCircuitError` treats as a circuit failure by type.

Measured with a fake client answering `400 hostUnreachable general SOCKS server failure ttlExpired`:

```
pre-fix : 3 POSTs, 11.0 s wall  (retried + NEWNYM path taken)
post-fix: 1 POST,   1.5 s wall  (isRetryableError false, isCircuitError false)
```

That is the new regression test, and it is red against the pre-fix `StateError('HTTP $status: $body')`.

Residual, pre-existing and out of scope: `TorHttpClient.readUtf8Body` (`:86-88`) has no size cap, on every call site including 2xx. Dropping the body from the error reduces what a peer can inject, not what it can make us allocate.

## Nitpicks

* **`markedDeletedForEveryoneFailed` untested** — fixed (`e65e097`). The new test also asserts the local tombstone still applies: local delete succeeds, propagation did not.
* **The group test passed with zero send targets** — confirmed and fixed (`e65e097`). With no `group_members` rows the per-target loop at `message_modify_service.dart:390-402` never ran and `allDelivered` stayed `true` by default. It now inserts members and installs a deterministic `SideChannelPostman`, plus a mixed-outcome test (one member delivered, one rejected → `markedDeletedForEveryoneFailed`). Proof it is no longer vacuous: removing the member rows flips that test to `markedDeletedForEveryone`, which is the wrong-reason pass.
* **Test doubles duplicated in two files** — declined. This suite has **zero** relative imports across its ~200 test files; private per-file doubles are the established convention (`_FakePostman` is defined in 10 files, `_FakeHttpClientResponse` in 4 — two of which, `test/security/group_profile_fetch_gate_test.dart` and `test/security/group_invite_mode_test.dart`, pre-date this PR and would not be deduplicated by the proposed helper anyway). Adding `test/support/` would introduce a second convention next to an existing one for 34 lines of test doubles.
* **`try/finally` instead of `addTearDown` for process-global state** — fixed (`dc033f7`, `677014e`). Every reset is now registered at the point of assignment in both files.
* **Exception branch of `WsFrameRouter` untested** — fixed (`e65e097`). Proven mordant: with the `try/catch` removed the new test fails with the error thrown out of `handleInboundFrame` instead of the error frame.

PR #144, opened 2026-08-09
fix(ui): keep the composer from overflowing with the keyboard open

MrBuondi:
## Problem

The device log reports `A RenderFlex overflowed by 18 pixels on the bottom` during a chat session
and 34 pixels shortly after startup.

The chat body is `Column [ Expanded(list), composer ]`. The composer is a non-flex child, so
RenderFlex lays it out with unbounded main-axis constraints and it keeps its full intrinsic height
(reply preview + typing bar + input row), while the keyboard inset shrinks the body. When the
composer is taller than the space left, the Column overflows.

Reproduced at 360x496 with a 300px inset: 144px composer into 126px of body.

## Change

- Constrain the composer to the available height and let it scroll.
- The wrapper lives in a single `PrysmConstrainedComposer` used by the 1:1, group and self chats,
  instead of being triplicated inline. This is what makes the fix testable: the overflow test now
  exercises the shipped widget, so reverting the fix fails the test. The previous test pumped a
  hand-rolled copy of the screen's body and stayed green when the fix was removed.
- The PIN screen gets the same treatment for the startup case.
- The assertion no longer pins an exact pixel count, which depended on font metrics.

## Verification

- Mordancy demonstrated: reverting the wrapper makes the test fail with the overflow error;
  restoring it passes.
- The three chat screens render correctly in the running Linux app, with
  `PrysmConstrainedComposer` present in the live widget tree and no overflow errors logged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented chat composers in direct, group, and self chats from overflowing when keyboards, reply previews, or typing indicators reduce available space.
  * Improved PIN entry layout on small screens by allowing content to scroll instead of overflowing.

* **Tests**
  * Added coverage for chat overflow scenarios and PIN entry layouts with biometric authentication and Tor progress elements.
  * Verified composer behavior under constrained screen sizes and keyboard insets.

### DIFF
--- a/lib/screens/chat.dart
+++ b/lib/screens/chat.dart
@@ -30,6 +30,7 @@ import 'package:prysm/database/messages.dart';
 import 'package:prysm/screens/chat_profile_screen.dart';
 import 'package:prysm/ui/chat/prysm_bubble_renderer.dart';
 import 'package:prysm/ui/chat/prysm_chat_composer_column.dart';
+import 'package:prysm/ui/chat/prysm_constrained_composer.dart';
 import 'package:prysm/ui/chat/prysm_chat_list.dart';
 import 'package:prysm/ui/chat/chat_search_bar.dart';
 import 'package:prysm/ui/chat/prysm_date_header.dart';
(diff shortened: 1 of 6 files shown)


MrBuondi:
@coderabbitai review

xmreur:
@coderabbitai review

MrBuondi:
Real, and fixed in `6d71021`: `UnlockLockoutService.setUseInMemoryStorageOnly(true)` is process-global state that the test never restored, so the mode was left on for whatever ran after it. The reset is now registered with `addTearDown` at the point of the assignment, next to the existing `tester.view.reset`.

One qualification, so the record is accurate: the leak is not reachable today. `flutter test` runs each test file in its own isolate and this is the last test in the file, so nothing observes the flag afterwards. What the fix removes is the ordering dependency — the test stops being correct only because of where it happens to sit in the file. The repo already resets the flag this way in `test/unlock_lockout_service_test.dart` and `test/unlock_flow_characterization_test.dart`.

Gate on the branch: `flutter analyze` 0 issues, `flutter test` **1003 passed + 1 skip** (unchanged; the fix adds no test).

xmreur:
build:test

xmreur:
## PR build results

|---|---|---|
| Windows | success | [Download](https://github.com/xmreur/prysm/actions/runs/31401380036/artifacts/9068064895) |
| Linux | success | [Download](https://github.com/xmreur/prysm/actions/runs/31401380036/artifacts/9067881189) |
| macOS | success | [Download](https://github.com/xmreur/prysm/actions/runs/31401380036/artifacts/9068012746) |
| iOS | success | [Download](https://github.com/xmreur/prysm/actions/runs/31401380036/artifacts/9068096610) |
| Android | success | [Download](https://github.com/xmreur/prysm/actions/runs/31401380036/artifacts/9068251867) |

Triggered by: @xmreur
Run: https://github.com/xmreur/prysm/actions/runs/31401380036

xmreur:
build:test

xmreur:
## PR build results

|---|---|---|
| Windows | failure | N/A |
| Linux | failure | N/A |
| macOS | failure | N/A |
| iOS | failure | N/A |
| Android | failure | N/A |

Triggered by: @xmreur
Run: https://github.com/xmreur/prysm/actions/runs/31404585670

PR #150, opened 2026-08-12
fix(transfer): agree on the file envelope scheme so chunked sends can start

MrBuondi:
Found by measuring, not by reading: a two-peer transmission campaign over real Tor reported `path=mono` for every attachment, including an 8 MiB one. The sender's log said why, identically on every attempt:

```
ERROR [ChatService]: Chunked file send failed: FormatException: Invalid file envelope
ERROR [ChatService]: Chunked file transfer failed for cf291efe-…, falling back to HTTP send
```

**Cause.** `parseFileTransferParts` (`file_transfer_sender.dart:28`) accepted only `file-aead-1`, but `792d98f` (2026-08-04) changed `CryptoWire.encryptFile` to hand the peer a **`file-signed-1`** envelope whose `wrappedKey` is a signed `dm-signed-2` map. The two shapes are otherwise identical — `crypto: v2`, `wrappedKey` map, base64 `nonce`, base64 `ciphertext` — so the whole defect is one string, and it throws on the first line of `FileTransferSender.send`, before a byte moves. `file-aead-1` survives only as the *self* payload, which never leaves the device. At `4922f28` all three sides agreed; `792d98f` changed the producer and touched neither the parser nor the receiver. **The chunked path has been dead for eight days.**

A second, independent disagreement sat behind it: `handleEnd` (`file_transfer_handler.dart:321`) rebuilt every transfer as `file-aead-1`, so a signed wrap re-tagged unsigned would have been refused by the inbound auth gate (`direct_message_auth.dart:191`) and then by `decryptLegacyFromPeer`, which knows only `dh-aead-1/2`. Fixing only the parser would have moved the failure, not removed it.

**Fix.** The parser accepts both schemes; the sender puts the one it parsed into `file_transfer_begin`; `handleEnd` rebuilds with that scheme. A begin frame from a peer predating the field — or carrying anything other than the two known constants — is treated exactly as before, `file-aead-1`, so an older peer still interoperates and no attacker-chosen string reaches the crypto layer. `encryptFile` keeps signing peer payloads and `allowLegacyUnsignedFile` stays off: authentication is not relaxed to make the transport agree.

**Why the suite was green.** `file_transfer_sender_test.dart` hand-built a `file-aead-1` envelope instead of calling `encryptFile`, so it asserted the parser's own stale assumption and never crossed the seam. The new tests build the payload the way production does and carry it end to end: real `encryptFile` output → `FileTransferSender.send` → the captured begin payload → real `handleBegin` → real chunk frames → `handleEnd` → `decryptFileFromPeer` returns the original bytes. Plus the receiver rebuild in both directions (scheme present → `file-signed-1`, scheme absent → `file-aead-1`). Each was checked mordant: parser reverted → only the real-payload tests fail; `handleEnd` hardcoded again → only the rebuild test fails; legacy fallback dropped → only the legacy test fails.

**Verification.** `flutter analyze` 0, full suite 1062 + 1 skip green on this branch in isolation. Live between two containers with their own identities and Tor, 1:1 chat, same rig before and after:

| payload | before | after |
|---|---|---|
| 2 MiB | `path=mono`, 13.4 s, 1255 kbps, fallback error | `path=chunked`, **9 chunks**, 23.0 s, 729 kbps |
| 8 MiB | `path=mono`, 45.9 s, 1462 kbps, fallback error | `path=chunked`, **33 chunks**, 112.0 s, 599 kbps |
| 200 KiB | `path=mono`, 3.4 s | unchanged (below the 1 MiB threshold) |
| 997-char text | recv 0.68 s | recv 0.71 s, unchanged |

Sender now logs `chunk 33/33 sent bytes=16` then `transfer complete message=…`; receiver logs `begin accepted transfer=… chunks=33` and `Received file from …`. No fallback line, no failure signatures on either peer.

**One thing this fix exposes, and deliberately does not change.** Now that chunking actually runs, it is **1.7–2.4× slower** than the monolithic POST at these sizes: each 256 KiB chunk waits for its own ack, so 8 MiB is 33 sequential Tor round trips (measured inter-chunk gaps 1.7–2.7 s, with occasional 8–41 s stalls). What chunking buys is per-chunk retry (3×), progress reporting, and no 33% base64 inflation on the wire. Two follow-ups worth deciding separately, with numbers rather than opinion:

1. **Pipeline the chunk sends.** A window of N chunks in flight would remove the serialization; the ack correlation is already per-chunk index. This is the change that would make chunked strictly better than monolithic.
2. **`FileTransferSender.send`'s `timeout` parameter (`:67`, default 5 minutes) is dead.** It is passed from `ChatService._sendOverTor` and never used: begin/end use a hardcoded 30 s and each chunk ack 60 s. So there is no whole-transfer deadline on this path — which is why this fix does *not* introduce a cliff for large files, but the parameter currently lies about a bound that does not exist.

Until (1) lands, a maintainer may reasonably prefer to raise `chunkThresholdBytes` (`file_transfer_policy.dart:6`) above 1 MiB so the slower path is used only where its resilience is worth it. That is a product knob, so it is not in this PR.

## Summary by CodeRabbit

* **New Features**
  * File transfers now support cryptographically signed envelopes alongside existing encrypted transfers.
  * Transfer security information is preserved throughout sending and receiving.
* **Bug Fixes**
  * Improved processing and reassembly of completed signed file transfers.
  * Transfers without explicit security information remain compatible with the legacy encryption path.
* **Tests**
  * Added coverage for signed transfers, legacy transfers, payload handling, and end-to-end reassembly.

### DIFF
--- a/lib/services/file_transfer_handler.dart
+++ b/lib/services/file_transfer_handler.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 import 'dart:convert';
 
 import 'package:flutter/foundation.dart';
+import 'package:prysm/crypto/constants.dart';
 import 'package:prysm/crypto/envelope.dart';
 import 'package:prysm/server/inbound_limits.dart';
 import 'package:prysm/server/inbound_message_router.dart';
@@ -27,6 +28,7 @@ class _InboundTransfer {
     required this.ciphertextSize,
     required this.totalChunks,
     required this.chunkSize,
+    required this.scheme,
     this.replyTo,
     this.viewOnce = false,
   }) : buffer = Uint8List(ciphertextSize),
@@ -45,6 +47,7 @@ class _InboundTransfer {
   final int ciphertextSize;
   final int totalChunks;
   final int chunkSize;
+  final String scheme;
   final String? replyTo;
   final bool viewOnce;
 
@@ -215,6 +218,14 @@ class FileTransferHandler {
     final transferId = payload['transferId'] as String;
     final nonce = base64Decode(payload['nonce'] as String);
 
+    // Peers predating the scheme field send no scheme; anything that is not a
+    // clean `file-signed-1` is treated exactly like today's legacy
+    // `file-aead-1` transfer, and only the two known constants are ever
+    // stored or handed to the crypto layer.
+    final scheme = payload['scheme'] == CryptoConstants.schemeFileSigned1
+        ? CryptoConstants.schemeFileSigned1
+        : CryptoConstants.schemeFileAead1;
+
     _active[transferId] = _InboundTransfer(
       messageId: payload['messageId'] as String,
       senderId: payload['senderId'] as String,
@@ -230,6 +241,7 @@ class FileTransferHandler {
       ciphertextSize: payload['ciphertextSize'] as int,
       totalChunks: payload['totalChunks'] as int,
       chunkSize: payload['chunkSize'] as int,
+      scheme: scheme,
       replyTo: payload['replyTo'] as String?,
       viewOnce: payload['viewOnce'] == true,
     );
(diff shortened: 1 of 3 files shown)


xmreur:
@coderabbitai review

PR #151, opened 2026-08-12
feat(lab): measure attachments and both ends of a send, not just arrival

MrBuondi:
`txlab` shipped in #149 could only type a short unique mark into the composer and tap send. Two whole classes of report were therefore unmeasurable: **articulated multi-paragraph messages** and **attachments** — and even for text it reported only when the message *arrived*, never when the sender was done.

This is the instrument that produced the numbers in #148 and #150. Landing it is what makes those numbers reproducible.

## What it does now

- **`send FROM TO --text TEXT | --text-file PATH`** — a real body. The mark has to be its **first** line: `prysmlab type` echoes the composer as `now=<value>` on one line, so a body with newlines is only visible up to the first one, and verifying against the whole body skipped every single rep (observed: three reps, three `SKIPPED`).
- **`file FROM TO --size 2MiB [--type image] [--reps N]`** — attachments, the only way a harness can send them: `ChatService.sendFileMessage` under `--lib package:prysm/screens/chat.dart`. The UI path is not drivable — `_handleSendImage`/`_handleSendFile` open native pickers and the image flow blocks on a modal preview screen. The payload is staged **inside** the sender's container through the existing eval channel (no new docker plumbing): a 64 KiB pseudo-random block repeated with native `writeFromSync`, because generating N bytes closure-by-closure stalls the app isolate. Every one of those evals resets the sticky `--lib` in a `finally` — a leftover `--lib` makes the next `tap`/`type` in that container die with a cryptic Dart compile error.
- **Both ends of every message.** Per rep: `send=` (sender's transport completion), `recv=` (receiver's arrival line), `render=` (bubble in the receiver's widget tree), all against one host clock taken immediately before the send; plus `chunks=`, `path=chunked|mono` and `kbps=` for attachments, and median/min/max in one shared summary.

Two things that look like harness bugs and are not, now documented in the skill: `send` larger than `recv` by ~0.3 s on the WS path is the ack being logged after delivery, and the sender's `send ok` can land *after* the receiver's bubble — it needs a bounded poll, not a single read, or the rep silently reports no send side (`path=?`, `ack=n/a`).

## Skill

Rule 18 records what the instrument measured: 997-char body recv **0.68 s**; 200 KiB attachment 2.1–3.4 s; 2 MiB monolithic 13.4 s (1255 kbps) vs **23.0 s** (729 kbps) chunked; 8 MiB monolithic 45.9 s (1462 kbps) vs **112.0 s** (599 kbps) chunked in 33 chunks; contact add 5.5 s each way; 0 lost in 16 messages / 11.4 MiB.

Rule 15 gets a **correction**, not an addition. It claimed the soft keyboard can never be raised on the headless AVD, "even for an explicit `TextInput.show`". That is true only for *injected* pointers: with real touch (`adb shell input tap`, coordinates in physical pixels) the IME comes up every time — measured on debug and release builds, `mInputShown` true → BACK → false → tap → true. The whole class of "the keyboard does not come back" defects is testable here, which is how the dead tap area in #145 was found and proven. Two traps are written down: a BACK sent while the keyboard is already closed pops the route and eventually tears the tree down, and the first real tap after a relaunch or a system permission dialog is swallowed.

## Verification

`python3 -m py_compile tool/live/txlab` silent; `--help` for `txlab`, `send` and `file` all print with the new surface; unit checks on the pure helpers (`parse_size` incl. junk rejection, the first-line mark composer, the composer-clean check accepting `now=<mark>` and rejecting anything else, `parse_log_ts` against a literal sender line). No `lib/` change, no dependency, nothing outside `tool/live/txlab` and the skill.

The end-to-end proof is the four campaigns it drove today against two live containers over real Tor: the before/after of #150 (mono → 9 and 33 chunks), the before/after of #148 (sidebar refresh 22.45 s → 0.68 s), and the text/attachment baselines above.

## Summary by CodeRabbit

* **New Features**
  * Added support for sending UTF-8 text files and file/image attachments.
  * Added configurable payload sizes, file types, repetitions, and timeouts.
  * Added transfer timing summaries for sending, receiving, rendering, acknowledgements, and chunking.
  * Added multi-line text support with complete composer-content validation.
* **Bug Fixes**
  * Improved handling of delayed acknowledgements.
  * Improved detection of attachment rendering and transfer completion.
* **Documentation**
  * Expanded soft-keyboard testing guidance for real touch input, focus handling, visibility checks, and layout changes.

### DIFF
--- a/.agents/skills/live-app-testing/SKILL.md
+++ b/.agents/skills/live-app-testing/SKILL.md
@@ -228,11 +228,22 @@ are forwarded into the container.
     never has it. A missing "Paste" entry in the selection menu is this, not a defect. What is
     observable in-app is `Cut` — it mutates the field — and the app's own
     "Copied to clipboard" toast.
-    The **soft keyboard has the same cause and the same answer**: `show_ime_with_hard_keyboard
-    1` is accepted, the IME binds (`mHaveConnection=true`), and it still never shows
-    (`mInputShown=false`) even for an explicit `TextInput.show` — a windowless app cannot raise
-    it, so `viewInsets.bottom` is always 0 and no amount of tapping produces a keyboard-open
-    layout. To exercise what a keyboard actually does to layout, shrink the display instead:
+    The **soft keyboard is a different story, and the earlier version of this rule was wrong**:
+    it never shows for *injected* pointers (`tap`, `longpress`, `TextInput.show` from an `eval`)
+    — `mInputShown=false` no matter what — because Flutter's own pointer events never make the
+    Android view the IME's served view. It shows perfectly for **real touch**:
+    `adb shell input tap <x*dpr> <y*dpr>`, read back with
+    `adb shell dumpsys input_method | grep mInputShown`. Measured on the headless AVD, debug and
+    release builds alike: tap the field -> `mInputShown=true`, BACK -> `false`, tap again ->
+    `true`. So the whole class of "the keyboard does not come back" defects IS testable here —
+    it just needs `docker exec <container> adb shell input`, coordinates in physical pixels
+    (logical × `devicePixelRatio`, 2.75 on the `pixel_5` AVD), and widget geometry read from the
+    tree with an `eval` because a screenshot cannot tell you where a 20 dp text strip ends.
+    Two traps that cost a session: a BACK sent while the keyboard is already closed pops the
+    route and eventually tears the app's tree down (guard every BACK on `mInputShown=true`), and
+    the first real tap after an app relaunch or a system permission dialog is swallowed — check
+    `adb shell dumpsys window | grep mCurrentFocus` before trusting a negative result.
+    To exercise what a keyboard does to *layout* without the keyboard, shrink the display:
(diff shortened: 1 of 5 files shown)


MrBuondi:
All six findings are addressed as of `cae06c4`; nothing outstanding on this branch.

For the record, since only five carry the bot's auto-marker: the terminal-newline finding (`tool/live/txlab:277`) landed too. `compose_wire_body` is now `return f"{mark}\n{body}"` with the `rstrip(chr(10))` gone, so a `--text`/`--text-file` body rides byte-for-byte, trailing newlines included, and the docstring says why.

The other five: `UnicodeError` caught alongside `OSError` on the `--text-file` read; `verify_composer_text` compares the **whole** `EditableTextState.textEditingValue.text` against the wire body rather than the first-line `now=` echo; `cmd_file` rejects `--size` above 50 MiB immediately after `parse_size`, before anything is staged; the `ImageMessageBubble` baseline is counted before `t0`; and the throughput metric is `payload_kbps` throughout, documented as plaintext-over-receive-time, not bytes on the wire.

MrBuondi:

PR #152, opened 2026-08-12
perf(transfer): keep eight chunks in flight instead of one

MrBuondi:
#150 made chunked transfers work again after eight days dead. Measuring the result showed the path was **slower than not using it**: 2 MiB took 23.0 s against 13.4 s monolithic, 8 MiB 112.0 s against 45.9 s. The sender awaited an ack after every 256 KiB, so 8 MiB was 33 sequential Tor round trips — measured inter-chunk gaps 1.7-2.7 s. The bytes were never the problem.

## The change

Eight chunks in flight instead of one. What is remarkable is how little had to move, because the pieces were already there:

- the ack frame already carries the chunk index (`file_transfer_chunk_ack` payload `{transferId, chunkIndex}`) and `pendingChunkAcks` was **already** keyed `transferId:index` — only the control flow serialised;
- the receiver already writes by index into a preallocated buffer, acks duplicates, and decides completeness by counting **distinct** indices, so out-of-order arrival needed no change at all;
- chunk acks mean "written to the in-memory buffer", not "persisted" — the single DB write stays in `handleEnd` — so a window does not change the disk pattern.

Per-index retries keep the old semantics: an ack timeout re-sends that index alone while the rest of the window keeps flying, and exhausting `maxChunkRetries` still fails the whole transfer with the same log line. `file_transfer_end` waits for **every** index rather than an in-order prefix, because `handleEnd` validates by count and would answer `Incomplete transfer`.

`send`'s `timeout` parameter also goes: it was passed from `ChatService._trySendChunkedFile` and never read in the body. The per-chunk 60 s × 3 retries plus the abort path are the real bound. Inventing a whole-transfer deadline to replace a dead one is a product decision, not a refactor — flagging it rather than deciding it.

## Measured, two containers, real Tor, same rig

| payload | sequential (#150) | windowed (this PR) |
|---|---|---|
| 2 MiB | 23.0 s · 729 kbps | **14.3 s · 1173 kbps** |
| 8 MiB | 112.0 s · 599 kbps | **41.2 s / 44.1 s · 1630 / 1523 kbps** |

41 s also beats the 45.9 s monolithic best for the same file, because chunks carry raw ciphertext where the monolithic envelope carries base64 (+33% on the wire). The window turns the slowest path into the fastest one.

**Why eight and not sixteen.** W=16 measured 79.5 s against 87.3 s for W=8 on a slow circuit — a 9% difference, while Tor circuit variance across today's runs was 1.7-9× (the same 8 MiB monolithic send took 45.9 s on one circuit and 416 s on another; 200 KiB took 3.4 s and 150 s). That difference is not distinguishable from noise, so the smaller window wins: it collapses the stall waves from 33 to 5 and buffers 2 MiB of unacked frames instead of 4. `WebSocket.add` gives no backpressure, so the window size *is* the sender's buffer bound.

## Tests

The existing suite could not see this change: its fake link acks **synchronously inside `sendBytes`**, so sequential and windowed behave identically under it. Five new tests use a gated link that records frames and acks only when released:

- eight frames on the wire before any ack is released, never more in flight;
- releasing one ack lets exactly one more chunk out;
- withholding one middle ack re-sends only that index and the transfer still completes;
- `file_transfer_end` is not requested until every index is acked;
- an index that is never acked fails the send after `maxChunkRetries`.

Each was checked mordant: reverting to the sequential driver fails the first, second and fourth; making a retry re-send the whole window fails the third; removing the all-acks-before-end wait fails the fourth.

`flutter analyze` 0, full suite **1094 + 1 skip** green on this branch in isolation.

## Two findings from the measurement, not fixed here

1. **The first attachment after a (re)connect always takes the monolithic path**, because `shouldUseChunkedTransfer` requires an already-established WS link. On a cold circuit that is the worst possible choice: measured 150 s for a 200 KiB file and 416 s for 8 MiB, immediately after a restart. Warming the link (or letting the chunked path bring it up, as `prepareForFileTransfer` already can) would matter more than any window tuning.
2. That 416 s monolithic send **exceeded the 5-minute cap** `ChatService` puts on `postDirect` and still arrived, which means it went through the pending-queue retry path. Worth a look on its own: the cap does not abort the delivery, it just re-runs it later.

One test seam was added: `FileTransferSender.ackTimeout`, a constructor field defaulting to the existing 60 s, because the retry and abort tests cannot use `fake_async` — it is only a transitive dependency of `flutter_test` (importing it would fail `depend_on_referenced_packages`, and adding it to the pubspec would break the no-new-dependencies rule), and the sender's `finally` awaits `controlSub.cancel()`, whose continuation is scheduled on the real microtask queue and never resumes under fake time.

## Summary by CodeRabbit

* **Improvements**
  * File transfers now send multiple chunks concurrently for improved throughput.
  * Transfers continue progressing as chunks are acknowledged, with bounded buffering for consistent performance.
  * Individual unacknowledged chunks are retried without restarting the entire transfer.
* **Reliability**
  * Transfers wait for all chunks to be confirmed before completing.
  * Failed transfers are reported clearly after retry limits are reached.
  * Send failures no longer leave transfers retrying after cancellation.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -12,15 +12,10 @@ concurrency:
   group: build-pr-${{ github.event.issue.number }}
   cancel-in-progress: true
 
+# Pinned, not tracked: an unpinned `stable` moved 3.44.9 -> 3.47.0 under an
+# unchanged main and broke it. Matches both Containerfiles. Bump deliberately.
 env:
   PR_NUMBER: ${{ github.event.issue.number }}
-
-# The toolchain is pinned, not tracked. `channel: stable` alone put CI on
-# whatever Flutter shipped that week: 3.44.9 was green on d18d384 and 3.47.0
-# failed the identical commit hours later, on a new lint plus a dart:io
-# signature change. Both Containerfiles already pin 3.44.8 (host toolchain);
-# this is the environment that was drifting away from them. Bump deliberately.
-env:
   FLUTTER_VERSION: "3.44.8"
 
 jobs:
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -15,6 +15,14 @@ concurrency:
 env:
   PR_NUMBER: ${{ github.event.issue.number }}
 
+# The toolchain is pinned, not tracked. `channel: stable` alone put CI on
+# whatever Flutter shipped that week: 3.44.9 was green on d18d384 and 3.47.0
+# failed the identical commit hours later, on a new lint plus a dart:io
+# signature change. Both Containerfiles already pin 3.44.8 (host toolchain);
+# this is the environment that was drifting away from them. Bump deliberately.
+env:
+  FLUTTER_VERSION: "3.44.8"
+
 jobs:
   authorize:
     if: ${{ github.event.issue.pull_request && github.event.comment.body == 'build:test' }}
@@ -77,6 +85,7 @@ jobs:
 
       - uses: subosito/flutter-action@v2
         with:
+          flutter-version: ${{ env.FLUTTER_VERSION }}
           channel: stable
           cache: true
 
@@ -112,6 +121,7 @@ jobs:
 
       - uses: subosito/flutter-action@v2
         with:
+          flutter-version: ${{ env.FLUTTER_VERSION }}
           channel: stable
           cache: true
 
@@ -156,6 +166,7 @@ jobs:
 
       - uses: subosito/flutter-action@v2
         with:
+          flutter-version: ${{ env.FLUTTER_VERSION }}
           channel: stable
           cache: true
 
@@ -198,6 +209,7 @@ jobs:
 
       - uses: subosito/flutter-action@v2
         with:
+          flutter-version: ${{ env.FLUTTER_VERSION }}
           channel: stable
           cache: true
 
@@ -242,6 +254,7 @@ jobs:
 
       - uses: subosito/flutter-action@v2
         with:
+          flutter-version: ${{ env.FLUTTER_VERSION }}
           channel: stable
           cache: true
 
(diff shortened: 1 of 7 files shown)


MrBuondi:
Both remaining findings addressed in `323bf79`.

**Abandoned chunk loops (`file_transfer_sender.dart` 262-283).** Applied as proposed: an `abandoned` local, set in the `finally` before the completers are errored, checked at the top of every attempt alongside `transferFailed.isCompleted`. It stays a local of `send` — a field would leak one send's abandonment into a concurrent one.

**No test ships with it, deliberately.** The guard's trigger is a microtask-ordering race: the `finally`'s continuation has to interleave between an ack timing out and its `catch` resuming. Everywhere else the `finally`'s `completeError` already kills the waiting chunk. I could not reproduce it — with the guard removed, the candidate test passed at 30 ms, 2 ms and 700 us ack timeouts (the only failures at 700 us were the harness's own polling race, present with and without the guard). This repo checks every new test mordant, and a test that passes with the fix removed is worse than no test: it claims coverage it does not have. The guard is kept as cheap defence, not as covered behaviour.

**Fake-link duplication (`file_transfer_sender_test.dart` 98-161).** Extracted `_AckLinkBase`; `_ChunkAckLink` and `_GatedAckLink` now override only `sendBytes`. Public surface unchanged, no test body touched, **44 lines removed**.

`flutter analyze` 0, full suite **1095 + 1 skip** green on this branch in isolation.

Also opened, from the two findings this PR measured and left open plus the handoff's list:
- #154 — the cold-link monolithic fallback (finding 1 here)
- #155 — PIN keypad clipping + `onChanged` double-fire
- #153 — lab native-asset seeding

The 5-minute `postDirect` cap (finding 2) is untouched: it defers rather than aborts, and replacing it with a real deadline is a product decision, not a refactor.

xmreur:
@coderabbitai review

MrBuondi:

PR #153, opened 2026-08-13
tool(lab): seed the host's native assets so a failed download cannot kill the build

MrBuondi:
Operational, no `lib/` change.

`.dart_tool` is a staging exclude, so the build hook re-downloads `libsqlcipher.so` **inside the container every time**. When that download fails it leaves a **0-byte `.tmp`** behind and the whole build dies on `Building assets for package:sqlite3 failed`, with the lab unusable until the artifact is `docker cp`'d in by hand — and again after every `up --fresh`. That cost about 30 minutes in the last session.

`stage_repo` now copies whatever verified `.so` the host already has into the lab tree, where the hook finds it as a cache hit instead of a network call.

- The `download-*` directory name is a **version hash**, not a constant — this host holds four (`187fb732`, `92db9f41`, `aab3ca74`, `3b604415`). Globbed, not pinned.
- 0-byte sources are skipped, so a previously failed download is never propagated into the lab.
- Absent on the host is fine: the container downloads it exactly as it does now. No new failure mode.
- `--exclude .dart_tool` also keeps rsync's `--delete` off these, so **one seed survives every later `restart --sync`**.

Verified against the real host tree: `python3 -m py_compile` silent, and the seed walk copies all four artifacts (including the `187fb732` one the manual workaround named) into a scratch `WORK` at the identical relative path.

## Summary by CodeRabbit

- **Reliability**
  - Standardized automated builds and continuous integration on Flutter 3.44.8 across Windows, Linux, macOS, iOS, and Android.
  - Improves consistency and predictability when validating and packaging releases.

- **Bug Fixes**
  - Improved handling of native assets during live development and staging.
  - Helps ensure required build artifacts remain available across synchronization steps.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -12,15 +12,10 @@ concurrency:
   group: build-pr-${{ github.event.issue.number }}
   cancel-in-progress: true
 
+# Pinned, not tracked: an unpinned `stable` moved 3.44.9 -> 3.47.0 under an
+# unchanged main and broke it. Matches both Containerfiles. Bump deliberately.
 env:
   PR_NUMBER: ${{ github.event.issue.number }}
-
-# The toolchain is pinned, not tracked. `channel: stable` alone put CI on
-# whatever Flutter shipped that week: 3.44.9 was green on d18d384 and 3.47.0
-# failed the identical commit hours later, on a new lint plus a dart:io
-# signature change. Both Containerfiles already pin 3.44.8 (host toolchain);
-# this is the environment that was drifting away from them. Bump deliberately.
-env:
   FLUTTER_VERSION: "3.44.8"
 
 jobs:
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -15,6 +15,14 @@ concurrency:
 env:
   PR_NUMBER: ${{ github.event.issue.number }}
 
+# The toolchain is pinned, not tracked. `channel: stable` alone put CI on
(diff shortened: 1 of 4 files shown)


MrBuondi:
Both findings applied.

**Duplicate `env:`** — fixed in `9565920`, before this review round landed. One mapping, `PR_NUMBER` preserved.

**Atomic publish (`tool/live/prysmlab:688`)** — correct, and a sharp catch: the function guarded a 0-byte *source* and not a partial *destination*, which is the same failure it exists to prevent, made permanent. An interrupted `shutil.copy2` leaves a truncated `.so`; `os.path.exists(dst)` accepts it forever precisely because `--exclude .dart_tool` keeps rsync's `--delete` off it; every later run then hands the container a broken library rather than a missing one, which is strictly worse than the 0-byte `.tmp` this all started with.

Fixed in `87d3d5c`: copy to `dst + ".part"` and `os.replace` into place (same directory, so the rename is atomic), and compare **sizes** instead of mere existence so an already-damaged destination is repaired rather than skipped. Size is enough here — the source is a verified artifact and truncation is the failure mode; hashing 5 MB on every `up` would buy nothing.

Verified against the real host artifacts (4 of them on this machine):

| step | result |
|---|---|
| cold seed | 4 copied |
| second run | 0 copied (idempotent) |
| truncate one to 1/3 size, re-run | 1 copied, full size restored |
| `.part` residue | none |
| **previous code, same truncation** | **0 copied, file left broken** |

That last row is the mordant check: the old code makes the corruption permanent.

PR #154, opened 2026-08-13
fix(transfer): bring the link up before deciding against the chunked path

MrBuondi:
Closes the first of the two defects #152 found by measuring and did not fix.

The first attachment after a (re)connect **always** took the monolithic path. Measured immediately after a restart: **150 s for 200 KiB**, **416 s for 8 MiB**.

## Why forcing `wsConnected` would not have fixed it

`shouldUseChunkedTransfer` gates on three inputs, and **two** are link-derived — both read false on a cold link:

- `wsConnected` — the obvious one;
- `peerSupportsFileTransfer` — the capability map is **erased on every `_removeLink`** (`ws_connection_manager.dart:743`), so the capability is *unknowable* until the handshake has actually run.

The link has to be up before the predicate is evaluated. There is no flag that substitutes for it.

## The ordering was inverted

`FileTransferSender.send` **does** call `prepareForFileTransfer` — but the sender is only constructed at `chat_service.dart:707`, **inside the `if` that already lost the decision**. The bring-up sat strictly downstream of the decision that needed it, so on a cold link the decision short-circuited, the sender was never built, and the bring-up never ran.

## Why this is safe rather than a gamble

The monolithic fallback **already pays the identical connect cost and then throws the link away**. `postDirect` -> `postMessageOrFallback` calls `ensureConnected(connectBudget: interactiveConnectBudget)` — the same <= 25 s — then at `transport_provider.dart:436-444` refuses the WS it just built, because `shouldAvoidWsMonolithicSend` is true for >= 1 MiB media, and posts over HTTP.

So today a cold-link 8 MiB send pays <= 25 s of connect, **discards the link**, then spends 416 s on HTTP. The connect cost is already sunk. Moving it earlier is net-zero latency for >= 1 MiB and unlocks the path that #152 measured at 41 s.

This is why the change did not need a live Tor campaign to justify: the cost being moved is a cost already paid.

## The change

Caller-side ordering only. `shouldUseChunkedTransfer` keeps its signature and stays pure.

1. `TransportProvider.isConfigured` guard stays first.
2. **Size gate, hoisted out of the predicate.** This is the one real regression risk: without it every small attachment would pay up to 25 s (`interactiveConnectBudget`) for a path it will not take. `FileTransferPolicy.isChunkCandidate` keeps the threshold in one home, and the predicate uses it too.
3. Then the bring-up, in a `try`/`catch` that **logs** and returns false — the old monolithic path. It is bounded at 25 s and cannot hang, but it does mutate state: it pins the peer and clears backoff/quarantine. That is noted at the call site.
4. Then read `wsConnected` / `peerSupportsFileTransfer` fresh and evaluate exactly as before.

## Tests

Three decision-path tests against a **real** `ChatService` and `WsConnectionManager` — no new production test hooks were added.

| test | failure with the fix reverted |
|---|---|
| sub-threshold file never reaches the bring-up | `Expected: <0>` / `Actual: <1>` (size gate removed) |
| >= 1 MiB on a cold link brings the link up, then goes chunked | `Expected: <1>` / `Actual: <0>` (reordering reverted) |
| a throwing bring-up falls back without propagating | — |

The first also asserts a recorded connect failure **survives**: `pinPeer` is the first thing the bring-up does, so the backoff still being there proves the bring-up never ran.

`flutter analyze` 0, full suite **1092 + 1 skip** green on this branch in isolation.

## Still open

The 5-minute cap on `postDirect` does not abort delivery — a 416 s send blew through it and arrived anyway, via the pending-queue retry path. Unchanged here: the cap defers, it does not stop, and replacing it with a real deadline is a product decision.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file-transfer selection so smaller files use the standard HTTP transfer without unnecessary connection setup.
  * Large files now prepare the preferred transfer connection before switching to chunked delivery.
  * If connection preparation fails, transfers automatically fall back to HTTP instead of failing outright.
* **Reliability**
  * Added coverage for file-size thresholds, connection setup, and fallback behavior to help ensure consistent transfers.

### DIFF
--- a/lib/services/chat_service.dart
+++ b/lib/services/chat_service.dart
@@ -690,6 +690,18 @@ class ChatService {
     int? timestamp,
   }) async {
     if (!TransportProvider.isConfigured) return false;
+    if (!FileTransferPolicy.isChunkCandidate(fileSize)) return false;
+
+    // A cold link reads wsConnected=false and an empty capability map until
+    // the handshake runs, so bring the link up before evaluating the
+    // predicate; the monolithic fallback pays this same connect budget and
+    // discards the link. This pins the peer and clears its backoff/quarantine.
+    try {
+      await TransportProvider.instance.wsManager.prepareForFileTransfer(peerId);
+    } catch (e) {
+      Logging.error('Chunked link bring-up failed: $e', 'ChatService');
+      return false;
+    }
 
     final wsConnected = TransportProvider.instance.isRealtimeConnected(peerId);
     final peerSupports =
--- a/lib/util/file_transfer_policy.dart
+++ b/lib/util/file_transfer_policy.dart
@@ -24,12 +24,17 @@ class FileTransferPolicy {
   /// Per-chunk send retries before failing the transfer.
   static const int maxChunkRetries = 3;
 
+  /// True when [fileSizeBytes] reaches the chunked-transfer threshold.
+  /// Shared with the call-site size gate so the threshold has one home.
+  static bool isChunkCandidate(int fileSizeBytes) =>
+      fileSizeBytes >= chunkThresholdBytes;
+
   static bool shouldUseChunkedTransfer({
     required int fileSizeBytes,
     required bool wsConnected,
     required bool peerSupportsFileTransfer,
   }) {
-    if (fileSizeBytes < chunkThresholdBytes) return false;
+    if (!isChunkCandidate(fileSizeBytes)) return false;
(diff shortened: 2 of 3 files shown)


xmreur:
@coderabbitai review

MrBuondi:

xmreur:
@MrBuondi fix the conflict

PR #155, opened 2026-08-13
fix(ui): size the PIN keypad to the screen and stop onChanged firing twice

MrBuondi:
Two defects carried in the handoff's open list, both in widgets #145 had to leave alone. They land together because #145 is now in `main`.

## The `0` key is cut off, and no exception says so

`PinKeypad` is a hard **400 dp** block: four rows of 80 dp keys with 10 dp padding, no `LayoutBuilder`, no `AspectRatio`, nothing flexible. It cannot shrink. In onboarding step 2/7 the step viewport measures ~631 dp against ~736 dp of content, so the last row `['', '0', 'back']` falls below the fold.

The existing `renderflex_overflow_test.dart` style — `expect(tester.takeException(), isNull)` — is **blind to this**: the onboarding step scrolls, so nothing throws. The regression test asserts geometry instead, the `0` key's bottom edge against the viewport at scroll offset 0.

The key size now derives from `MediaQuery.sizeOf(context).height`, clamped to `[52, 80]`, `gap = keySize / 8`. `LayoutBuilder` will not work here: inside the step's `SingleChildScrollView` the incoming main-axis constraint is unbounded. At >= 1152 dp it reproduces 80 dp / 10 dp exactly, so the widget is byte-identical where it already fit; the 52 dp floor keeps every key above the 48 dp touch-target minimum. Fonts scale with the key.

**It also fixes a site nobody had looked at.** `PinPadScreen` (Settings -> Change passcode, Panic PIN) hosts the same keypad in a plain **non-scrollable** `Column` and throws a real `RenderFlex overflowed by 29 pixels` on a 360x600 phone. #144 fixed `pin_entry.dart` and never touched it. One file, three hosts, zero call-site churn.

One correction to the handoff: `PinPadScreen` does **not** overflow at 1080x2340 (content ~572 dp in ~724 dp available). It starts overflowing at logical heights <= ~615 dp, so that test uses 1080x1800 @3x — the shortest real phone class that fails before the fix.

## `onChanged` fires twice per keystroke, and once for a caret move

`_onTextChanged` is registered as a controller listener **and** `onChanged` is passed down to the `EditableText`. Inside `_formatAndSetValue`, `_value = value` writes `widget.controller.value`, and the controller is a `ValueNotifier` — the listener fires, then fourteen lines later the `EditableText` callback fires. Deleting the listener's call leaves `PrysmTextField` matching `PrysmSearchField`, which already only calls `setState`.

The doubling is the smaller half. The listener path is **unguarded**, where the `EditableText` path is gated on `oldValue.text != currentText`. A `TextEditingController` also notifies on **selection** changes — so moving the caret in the composer invoked `onChanged` -> `_notifyTypingFromText` and told the peer we were typing with **nothing typed**. The sharper of the two new tests pins that at zero.

The `setState` stays: `build` reads `controller.text.isEmpty` for the hint.

## Verification

`flutter analyze` 0, full suite **1095 + 1 skip** green on this branch in isolation (`main` is 1089 + 1).

Every new test checked mordant, with the production fix reverted:

| test | failure with the fix removed |
|---|---|
| `0` key above the fold | `Expected: a value less than or equal to <712.0>` / `Actual: <759.0>` |
| `PinPadScreen` no overflow | `Actual: FlutterError:<A RenderFlex overflowed by 29 pixels on the bottom.>` |
| one keystroke -> one `onChanged` | `Expected: <1>` / `Actual: <3>` |
| caret move -> zero `onChanged` | `Expected: <0>` / `Actual: <3>` |

The third test pins the no-op: 80 px keys / 10 px gaps / 400 dp / 32 pt digits on a tall surface.

## Summary by CodeRabbit

* **New Features**
  * PIN keypads now adapt key sizes and spacing to available screen height.
  * Key labels and controls scale consistently across screen sizes.

* **Bug Fixes**
  * Prevented duplicate text-change notifications when only the caret moves.
  * Improved short-screen PIN layouts to keep controls visible and avoid overflow.
  * Text field hints now disappear correctly after text entry.

* **Tests**
  * Added coverage for responsive keypad dimensions, short-screen layouts, PIN visibility, and text-field change behavior.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -12,15 +12,10 @@ concurrency:
   group: build-pr-${{ github.event.issue.number }}
   cancel-in-progress: true
 
+# Pinned, not tracked: an unpinned `stable` moved 3.44.9 -> 3.47.0 under an
+# unchanged main and broke it. Matches both Containerfiles. Bump deliberately.
 env:
   PR_NUMBER: ${{ github.event.issue.number }}
-
-# The toolchain is pinned, not tracked. `channel: stable` alone put CI on
-# whatever Flutter shipped that week: 3.44.9 was green on d18d384 and 3.47.0
-# failed the identical commit hours later, on a new lint plus a dart:io
-# signature change. Both Containerfiles already pin 3.44.8 (host toolchain);
-# this is the environment that was drifting away from them. Bump deliberately.
-env:
   FLUTTER_VERSION: "3.44.8"
 
 jobs:
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -15,6 +15,14 @@ concurrency:
 env:
   PR_NUMBER: ${{ github.event.issue.number }}
 
+# The toolchain is pinned, not tracked. `channel: stable` alone put CI on
+# whatever Flutter shipped that week: 3.44.9 was green on d18d384 and 3.47.0
+# failed the identical commit hours later, on a new lint plus a dart:io
+# signature change. Both Containerfiles already pin 3.44.8 (host toolchain);
+# this is the environment that was drifting away from them. Bump deliberately.
+env:
+  FLUTTER_VERSION: "3.44.8"
+
 jobs:
   authorize:
     if: ${{ github.event.issue.pull_request && github.event.comment.body == 'build:test' }}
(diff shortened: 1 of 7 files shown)


xmreur:
@coderabbitai review

MrBuondi:

PR #161, opened 2026-08-14
fix(inbound): gate unknown senders on an opt-in setting, and block peers in groups too

### LINKED ISSUE #159: Blocked non-contacts can still deliver messages after a delay
## Summary

When **refuse messages from non-contacts** is enabled, a non-contact can still eventually deliver a message. Delivery is delayed by roughly 3–4 minutes rather than being rejected.

This makes the privacy setting ineffective: the sender is not a contact, `/sync-hint` initially returns `403`, but the message is later accepted and routed.

## Expected behavior

When the setting to refuse messages from non-contacts is enabled, messages from users who are not contacts should be rejected consistently. They should not be delivered later through retry, sync, WebSocket, or fallback transport paths.

## Actual behavior

A message from a non-contact is eventually received after several minutes. The log shows repeated `403` responses for `/sync-hint`, followed by successful message delivery and routing.

## Relevant log excerpt

```text
[2026-08-14 14:47:40.300] INFO [PrysmServer]: POST - sync-hint
[2026-08-14 14:47:40.307] INFO [PrysmServer]: 2026-08-14T14:47:40.300318  0:00:00.007322 POST     /sync-hint
[2026-08-14 14:48:05.569] INFO [PrysmServer]: POST - sync-hint
[2026-08-14 14:48:05.588] INFO [PrysmServer]: 2026-08-14T14:48:05.569861  0:00:00.018360 POST     /sync-hint
[2026-08-14 14:48:30.216] INFO [PrysmServer]: POST - sync-hint
[2026-08-14 14:48:30.219] INFO [PrysmServer]: 2026-08-14T14:48:30.216464  0:00:00.002801 POST     /sync-hint
[2026-08-14 14:48:55.176] INFO [PrysmServer]: POST - message
[2026-08-14 14:48:55.302] INFO [InboundMessageRouter]: Received text from dpdhp5…
[2026-08-14 14:48:55.393] INFO [PrysmServer]: POST - sync-hint
[2026-08-14 14:49:03.540] INFO [PrysmServer]: 2026-08-14T14:48:55.393495  0:00:08.146542 POST     /sync-hint
[2026-08-14 14:49:04.395] INFO [PrysmServer]: 2026-08-14T14:48:55.176627  0:00:09.218437 POST     /message
```

## Steps to reproduce

1. Enable the setting that refuses messages from non-contacts for User A.
2. Ensure User B is not in User A’s contacts.
3. From User B, send a message to User A.
4. Wait several minutes.
5. Observe that User A receives the message despite the setting being enabled.

## Possible area to investigate

The contact-policy check may be applied to the initial `/sync-hint` request but not consistently enforced on a later delivery route, retry, WebSocket connection, or HTTP fallback.

## Product/design consideration

Defaulting to refusing messages from non-contacts may also be counterintuitive.

For example, a user may publish their Prysm ID on X expecting others to be able to contact them. If non-contact messages are refused by default, people cannot reach them and neither party receives a clear explanation of why.

Possible improvements:

- Default to allowing messages from non-contacts, with an explicit opt-in privacy setting to refuse them.
- If refusing remains the default, make it prominent during onboarding.
- Provide clear feedback to both sender and recipient when a non-contact message is blocked.

MrBuondi:
Closes #159. The [diagnosis comment](https://github.com/xmreur/prysm/issues/159#issuecomment-5294225079) has the full trace; the short version is that the setting the issue assumes **did not exist**, and the `403 'Unknown sender'` came from a gate that only the hint path has.

## What was actually broken

| | before | after |
|---|---|---|
| `/sync-hint` from an unknown sender | `403` (unconditional, `inbound_message_router.dart:152-155`) | unchanged |
| DM from an unknown sender | **accepted and stored**, then `ensureUserExist` created their `users` row so the hint gate stopped 403-ing | refused when `refuseUnknownSenders` is on, and no `users` row is created |
| **group** message from a **blocked** peer | **stored and notified** — `_isBlockedDm` returned false whenever `groupId != null` | refused, like their DMs, calls, typing and file transfers already were |

The second row is the issue; the third is a hole found while reading it, and it needed no policy decision — blocking was enforced everywhere except that one branch.

## The change

One gate, at the single convergence point of HTTP `POST /message`, both WS routes and file transfer. `_isBlockedDm` is gone, replaced by `_senderRefused` called at `processMessage:268` — **before** `_handleChatMessage`'s `ensureUserExist`, so a refused stranger leaves no trace:

```dart
if (BlockService.instance.isBlocked(senderId)) return true;   // groups too, now
if (data['groupId'] != null) return false;                    // co-members stay reachable
if (!settings.refuseUnknownSenders) return false;             // default off, before any DB hit
return (await DBHelper.getUserById(senderId)) == null;        // same "unknown" as the hint gate
```

Deliberate choices:
- **Default off.** A published Prysm ID stays reachable; privacy users opt in. The `!settings.refuseUnknownSenders` short-circuit sits ahead of the DB lookup, so the default path costs nothing.
- **Ack-and-drop with a fake `200`**, reusing what blocked DMs already do: it leaks nothing about block-vs-refuse-vs-offline, and it stops the sender's retry storm instead of feeding it (the 3-4 minutes in the issue are sender-side retry, one 25 s `interactiveConnectBudget` per cycle).
- **"Unknown" means `getUserById == null`**, the same definition the hint gate uses. One definition, not two.
- **Group invites untouched** — `groupInviteMode` still governs those.

Known consequence of dropping blocked peers' group messages: their group history has holes, and the sender-key anti-replay index is not advanced for the dropped messages. Both follow from the decision that blocking blocks.

## Tests

`test/security/refuse_unknown_senders_test.dart`, harness copied from `inbound_sender_auth_test.dart` + the `MessagesDb` wiring of `inbound_profile_fetch_gate_test.dart`. Assertions are `messages` row counts and the presence/absence of a `users` row — observable state, not mock calls. Each was re-run against its own mutant by hand:

| test | mutant | verbatim failure |
|---|---|---|
| setting on: unknown-sender DM is acked and dropped, no users row | contact lookup replaced by `return false` | `Expected: <0>` / `Actual: <1>` |
| setting off: first-contact DM from an unknown sender is stored | setting read inverted (default becomes refuse) | `Expected: <1>` / `Actual: <0>` |
| setting on: DM from a sender with a users row is stored | — guards the false-positive edge of the two above |
| blocked sender group message is acked and dropped | `if (data['groupId'] != null) return false;` restored ahead of the block check | `Expected: <0>` / `Actual: <1>` |

Plus one `testWidgets` in `test/group_invite_mode_screens_test.dart` for the privacy toggle, mordant against removing the row and against a repaint-only handler.

The four tests most likely to regress — `inbound_profile_fetch_gate_test.dart`, `group_profile_fetch_gate_test.dart`, `inbound_block_test.dart`, `inbound_sender_auth_test.dart`, all of which assert unknown-sender traffic **is** processed — stay green on the default.

## Verification

`flutter analyze` 0 · full suite **1129 + 1 skip** green on Flutter 3.44.8 (the pinned toolchain).

## Not in this PR

The issue's other two product asks: prominence during onboarding, and explicit feedback to the blocked sender. Feedback in particular cannot be done with the ack-and-drop response — telling the sender they were refused is exactly the leak that response avoids, so it needs its own decision.

## Summary by CodeRabbit

* **New Features**
  * Added a privacy setting to refuse direct messages from non-contacts.
  * Added a switch and explanatory text in Privacy Settings.
  * Blocked senders’ messages are acknowledged and discarded, while group messages remain unaffected by the non-contact setting.

* **Bug Fixes**
  * Blocked senders can no longer deliver messages through group conversations.
  * Invalid group identifiers on direct messages are rejected.

* **Tests**
  * Added coverage for the privacy setting, persistence, and message-filtering behavior.

### DIFF
--- a/lib/constants/group_constants.dart
+++ b/lib/constants/group_constants.dart
@@ -63,6 +63,20 @@ const Set<String> disappearingTimerTypes = {
   disappearingTimerType,
 };
 
+/// Every type whose payload legitimately carries a `groupId`. Inbound traffic
+/// that pairs a `groupId` with any other type is forging group scope, which
+/// would otherwise skip the direct-message authentication and the
+/// unknown-sender gate — both of which key off `groupId == null`.
+const Set<String> groupScopedTypes = {
+  ...groupControlTypes,
+  ...groupMessageTypes,
+  groupHistoryRelayType,
+  groupReactionType,
+  groupMessageModifyType,
+  groupReadReceiptType,
+  groupReadWaterlineType,
+};
+
 bool isGroupControlType(String type) => groupControlTypes.contains(type);
 bool isGroupMessageType(String type) => groupMessageTypes.contains(type);
 bool isReactionType(String type) => reactionTypes.contains(type);
@@ -70,6 +84,7 @@ bool isMessageModifyType(String type) => messageModifyTypes.contains(type);
 bool isReadReceiptType(String type) => readReceiptTypes.contains(type);
 bool isDisappearingTimerType(String type) =>
     disappearingTimerTypes.contains(type);
+bool isGroupScopedType(String type) => groupScopedTypes.contains(type);
 
 const Set<String> directMessageTypes = {
   'text',
--- a/lib/models/settings.dart
+++ b/lib/models/settings.dart
@@ -24,6 +24,10 @@ class Settings {
   final PanicAction panicAction;
   final GroupInviteMode groupInviteMode;
 
+  /// Refuse direct messages from senders with no `users` row. Off by default:
+  /// a published Prysm ID stays reachable unless the user opts out.
+  final bool refuseUnknownSenders;
+
   // Theme
   final int themeMode; // 0=light, 1=dark, 2=pink, 3=cyan, 4=purple, 5=orange
   final AppearanceSettings appearance;
@@ -58,6 +62,7 @@ class Settings {
     this.messageRetentionDays = 30,
     this.panicAction = PanicAction.decoy,
     this.groupInviteMode = GroupInviteMode.holdAsRequest,
+    this.refuseUnknownSenders = false,
     this.themeMode = 0,
     this.appearance = const AppearanceSettings(),
     this.avatar,
@@ -85,6 +90,7 @@ class Settings {
     'messageRetentionDays': messageRetentionDays,
     'panicAction': panicAction.name,
     'groupInviteMode': groupInviteMode.name,
+    'refuseUnknownSenders': refuseUnknownSenders,
     'themeMode': themeMode,
     'appearance': appearance.toJson(),
     'avatar': avatar,
@@ -114,6 +120,7 @@ class Settings {
     groupInviteMode: GroupInviteMode.fromJson(
       json['groupInviteMode'] as String?,
     ),
+    refuseUnknownSenders: json['refuseUnknownSenders'] ?? false,
     themeMode: json['themeMode'] ?? 0,
     appearance: AppearanceSettings.fromJson(
       json['appearance'] as Map<String, dynamic>?,
@@ -145,6 +152,7 @@ class Settings {
     int? messageRetentionDays,
     PanicAction? panicAction,
     GroupInviteMode? groupInviteMode,
+    bool? refuseUnknownSenders,
     int? themeMode,
     AppearanceSettings? appearance,
     String? avatar,
@@ -172,6 +180,7 @@ class Settings {
     messageRetentionDays: messageRetentionDays ?? this.messageRetentionDays,
     panicAction: panicAction ?? this.panicAction,
     groupInviteMode: groupInviteMode ?? this.groupInviteMode,
+    refuseUnknownSenders: refuseUnknownSenders ?? this.refuseUnknownSenders,
     themeMode: themeMode ?? this.themeMode,
     appearance: appearance ?? this.appearance,
     avatar: avatar ?? this.avatar,
@@ -215,6 +224,7 @@ class Settings {
         other.messageRetentionDays == messageRetentionDays &&
         other.panicAction == panicAction &&
         other.groupInviteMode == groupInviteMode &&
+        other.refuseUnknownSenders == refuseUnknownSenders &&
         other.themeMode == themeMode &&
         other.appearance == appearance &&
         other.avatar == avatar &&
@@ -242,6 +252,7 @@ class Settings {
         messageRetentionDays.hashCode ^
         panicAction.hashCode ^
         groupInviteMode.hashCode ^
+        refuseUnknownSenders.hashCode ^
         themeMode.hashCode ^
         appearance.hashCode ^
         (avatar?.hashCode ?? 0) ^
--- a/lib/screens/privacy_settings_screen.dart
+++ b/lib/screens/privacy_settings_screen.dart
@@ -40,6 +40,7 @@ class _PrivacySettingsScreenState extends State<PrivacySettingsScreen> {
   bool _typingIndicators = true;
   bool _lastSeen = true;
   bool _profilePhoto = true;
+  bool _refuseUnknownSenders = false;
   GroupInviteMode _groupInviteMode = SettingsService().groupInviteMode;
 
   /// True while a mode change is being persisted. Blocks a second selection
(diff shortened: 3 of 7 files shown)


MrBuondi:
The `groupId` finding is correct and is now fixed in `a295237` — and it was worse than the gate bypass it was filed against.

**Reproduced.** `validateMessage` enforces "group type ⇒ `groupId` present" (`inbound_message_router.dart:215-219`) and never the converse, so `{type: 'text', groupId: 'anything'}` passes validation. Two consequences, both confirmed by test:

1. The gate this PR added exempts group traffic on `data['groupId'] != null`, so a forged `groupId` skipped the unknown-sender check.
2. **Pre-existing on `main`:** `_handleChatMessage` authenticates only when `inboundGroupId == null` (`:552`), so the same forgery skipped `DirectMessageAuth` entirely — a stranger could store an unsigned direct message, and the `joinedAt` guard at `:542-550` does not stop it either (a non-member has `joinedAt == null`, so nothing drops).

**Fix, one allow-list instead of a per-callsite patch.** Classifying at each consumer would leave the next `groupId == null` branch to be found later, so the forgery is rejected at ingress: `groupScopedTypes` in `group_constants.dart` (spread of the existing `groupControlTypes` / `groupMessageTypes` sets plus the four group side-channel twins and `group_history_relay`), and one rule in `validateMessage`:

```dart
if (data['groupId'] != null && !isGroupScopedType(type)) {
  return InboundHandleResult.badRequest('groupId not allowed for a non-group type');
}
```

Allow-list, not deny-list, deliberately: a missing entry in a deny-list of direct types is a silent hole, while a missing entry here is a loud 400 on a legitimate feature. I checked the one type that pairs `groupId` with a non-group type in this codebase — `disappearing_timer_notice` — and it is a local `MessagesDb.insertMessage` (`disappearing_timer_service.dart:325-337`, `status: 'system'`), never inbound wire traffic.

**Regression test** as requested, `refuse_unknown_senders_test.dart`: unknown sender, `type: 'text'`, forged `groupId`, asserted `400` both with the setting on (the gate bypass) and off (the auth skip), plus no `messages` row and no `users` row. Mordant — deleting the rule gives `Expected: <400> / Actual: <200>`.

`flutter analyze` 0 · full suite **1130 + 1 skip** green.

MrBuondi:
@xmreur ready to merge from my side — please close it out. It closes #159.

Re-reviewed independently against `a295237`, with the two things an allow-list gate can get wrong checked explicitly:

- **Allow-list is complete and backward-compatible.** Every wire sender that attaches a `groupId`: group control channel (`group_invite`, `group_key_rotate`, `group_member_removed`, `group_profile_update`, `group_disappearing_timer`), `GroupChatService.postGroup` (`group_text`/`group_image`/`group_file`/`group_audio`), `side_channel_transport.dart:283` `_buildGroupPayload` (`group_reaction`, `group_read_receipt`, `group_read_waterline`, `group_message_modify`). All present in `groupScopedTypes`. Group typing is not a message type at all — it is the `typing_update` op (`typing_indicator_service.dart:152`), handled before `validateMessage`, so it cannot hit the new 400. File transfer and call payloads carry no `groupId` (`git grep -n groupId` over `file_transfer_sender.dart`, `file_transfer_handler.dart`, `call_service.dart` → no hits), and `sync-hint` payloads carry none either.
- **No version-skew break.** Each `group_*` side-channel type dates from the commit that introduced its own feature (`groupReactionType` 2e4e6d6, `groupReadReceiptType` af08da7, `groupMessageModifyType` 2e5b5f6, `groupReadWaterlineType` 3f903fa, `groupHistoryRelayType` 6f29b79), so no released peer ever paired a `groupId` with a direct type. The new 400 can only fire on forged traffic.
- **Ordering holds:** gate at `processMessage:274`, first `ensureUserExist` on the chat path at `:537`. Test 1 asserts the absence of the `users` row, not just the absence of the message.
- **Settings default is safe for existing installs:** `fromJson` reads `json['refuseUnknownSenders'] ?? false`, so a stored blob without the key neither throws nor flips to on; `SettingsService` is a singleton, so the router and the toggle share state.
- **Tests assert observable state** (message-row counts, `users` row presence, status codes), no mock-call assertions, and each has a distinct target line: `:934` (contact lookup), `:930-932` (setting short-circuit), `:922-924` (block check ahead of the groupId exemption), `:221-224` (forged-scope 400).

CI green on `a295237`, CodeRabbit's one finding resolved, no open threads.

Two things I found while reviewing that are **out of this PR's contract** and predate it — filing them separately rather than widening the diff: the call surface enforces `isBlocked` but not the new setting, and inbound group traffic never verifies the *sender's* membership.

PR #162, opened 2026-08-14
tool(lab): label a chunked transfer that fell back to HTTP

MrBuondi:
Found while measuring real send/receive times on the two-peer lab. No `lib/` change.

## The instrument was lying

`txlab file` decided the transport label from the sender's `begin transfer=` line alone. A chunked transfer that starts and then **dies on its ack timeout** is delivered by `ChatService`'s monolithic HTTP fallback, and the rep still printed `path=chunked`. That is how a cold-link rep read as

```
rep 1: send=n/a recv=220.431s render=n/a bytes=2097152 chunks=9 path=chunked payload_kbps=76
```

when the sender's own log says the chunked attempt was already dead:

```
16:38:39.312 begin transfer=eaa00554… chunks=9 ciphertext=2097168 peer=pvzcxv…
16:39:09.323 ERROR [FileTransferSender]: file transfer failed: TimeoutException after 0:00:30
16:39:09.332 ERROR [ChatService]: Chunked file transfer failed …, falling back to HTTP send
```

Now the rep prints `path=chunked->http` and takes its `send=` from `send ok` instead of the never-logged `transfer complete`. **Verified mordant** by patching `ackTimeout` to 1 ms in the staged copy and re-running a 2 MiB send: `rep 1: send=19.124s recv=18.605s render=19.048s bytes=2097152 chunks=9 path=chunked->http`, with the fallback line present in the sender log. Without the fix that rep reads `chunked`.

## Rule 18's numbers predated the chunk window

Measured today, same rig, chunked-vs-monolithic in **one session** (mono half by returning `false` from `shouldUseChunkedTransfer` in the staged copy, so the ≥1 MiB media still refuses monolithic WS and goes HTTP exactly as before chunking existed):

| payload | chunked (window=8) | monolithic HTTP | rule 18 before |
|---|---|---|---|
| 2 MiB | **9.4 s** median, 1778 kbps (n=3) | 10.7 s median, 1570 kbps (n=3) | chunked 23.0 s vs mono 13.4 s |
| 8 MiB | 53.5 s and 122.4 s (n=2) | 39.2 s and 50.8 s (n=2) | chunked 112.0 s vs mono 45.9 s |

So #152's window is worth 2.4× at 2 MiB and 2.1× at 8 MiB, and the old conclusion ("chunked is 1.7-2.4× slower than one POST") no longer holds at 2 MiB — it ties there and still loses at 8 MiB. Tor variance dominates single reps (the same 8 MiB chunked transfer ran 53 s and 122 s an hour apart), which is now stated in the rule.

Also recorded: group attachments are not measurable through `txlab file` (it drives `ChatService.sendFileMessage` on a live `ChatScreen`; a `GroupChatScreen` answers `NOTFOUND ChatScreen`), and the group path never chunks — one monolithic POST per member, 2 MiB in 10.2 s and 25.2 s driven by hand.

## New rule 19 — the cold link after a receiver restart

First send to a peer that just restarted: text **18.2 / 20.9 / 39.1 / 51.1 s** (n=4) against 0.75 s warm, 200 KiB 11.4 s against 1.8-2.7 s. Two 8 s `wsIfConnected` timeouts plus circuit rebuild. Once, the sender logged `WS ready` and opened a chunked transfer while the receiver's log shows no `handshake complete` for another 3.5 minutes — 30 s burned on a begin ack that could not arrive, then HTTP at 220 s. Filed separately as #163.

Two lab hazards from the same campaign are written down too: `restart --sync` twice wedged a peer on `UnlockScreen` with `Zone error: Build scheduled during frame` (plain `restart` recovers), and one unreproduced sighting of a receiver's open chat pane not showing 7 arrived-and-stored messages until an app restart — marked explicitly as a sighting, not a known bug, since three deliberate attempts to reproduce it all rendered normally.

`prysmlab down --purge` on both peers reports clean: no container, volume, staging or image left, and the repo working tree untouched.

## Summary by CodeRabbit

* **New Features**
  * Transfer reporting now identifies chunked transfers that fall back to HTTP after acknowledgment failures and uses completed HTTP timing when available.
  * Benchmark results now distinguish pre-window and August 14, 2026 measurements across text, attachments, pairing, loss, throughput, and group transfers.

* **Documentation**
  * Added observations on post-restart latency, group attachments, fallback behavior, rendering hazards, and transfer performance.
  * Updated conclusions to reflect current measurements rather than a previously consistent slowdown estimate.

### DIFF
--- a/.agents/skills/live-app-testing/SKILL.md
+++ b/.agents/skills/live-app-testing/SKILL.md
@@ -317,11 +317,12 @@ are forwarded into the container.
     2 MiB **chunked 9.4 s** (1778 kbps, n=3) versus **mono 10.7 s** (1570 kbps, n=3) — parity;
     8 MiB chunked 53.5 s and 122.4 s versus mono 39.2 s and 50.8 s. Pairing 5.7-7.9 s. 0 lost in
     31 messages / 32 MiB.
-    So the window closed most of the gap the older numbers show — 2.4× faster at 2 MiB, 2.1× at
-    8 MiB — and chunked is no longer the obviously wrong choice: it ties one POST at 2 MiB and
-    still loses at 8 MiB. Tor variance dominates single reps (the same 8 MiB chunked transfer ran
-    53 s and 122 s an hour apart), so never conclude from n=1, and re-measure both halves in ONE
-    session before touching either path.
+    So the window closed the 2 MiB gap outright — **2.4×** faster there (9.4 s median n=3 against
+    23.0 s) and chunked now ties one POST. At 8 MiB the two samples are 53.5 s and 122.4 s against
+    112.0 s before: **best case 2.1×, median 88 s ≈ 1.3×, n=2 — inconclusive, and the honest read
+    is that Tor variance is larger than the effect at that size.** Never conclude from n=1, quote
+    the aggregation with every headline, and re-measure both halves in ONE session before touching
+    either path.
     Group attachments are NOT in the harness: `txlab file` drives `ChatService.sendFileMessage` on
     a live `ChatScreen`, and a `GroupChatScreen` gives `NOTFOUND ChatScreen`. Driving
     `GroupChatService.sendFileMessage` by hand shows the group path never chunks (`path=mono`,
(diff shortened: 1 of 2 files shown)


MrBuondi:
Both correct, fixed in `dafb168` — and this is the same discipline the rule itself preaches, so worth being explicit about what was wrong.

**The 8 MiB claim was cherry-picked.** I wrote "2.1× at 8 MiB" from the 53.5 s sample and ignored my own 122.4 s one. Honest read now in the rule: samples 53.5 s and 122.4 s against 112.0 s before the window — best case 2.1×, **median 88 s ≈ 1.3×, n=2, inconclusive**, and at that size Tor variance is larger than the effect. Only the 2 MiB figure survives as a headline (9.4 s median, n=3, against 23.0 s → 2.4×), because it has an aggregation behind it.

**The 20-50× blended two different measurements.** Text and a 200 KiB attachment do not share a multiplier: the four cold text samples are 24-68× their warm median (18.2/20.9/39.1/51.1 s against 0.75 s), the 200 KiB one is ~4-6× (11.4 s against 1.8-2.7 s). The rule now scopes the headline to first *text* and reports the attachment separately.

Rule 18 also gained the general form of the lesson — quote the aggregation with every headline — since the reason my numbers misled is that a bare multiplier hides n.

MrBuondi:
@xmreur ready to merge from my side — please close it out.

One new commit since your last look: **`638bbef`** quotes `n` on the last three figures that lacked it, which is the same class as CodeRabbit's finding on `dafb168`:

- Baseline A is now labelled `single samples, n unrecorded per figure` (those numbers predate this campaign; inventing an `n` for them would have been worse than admitting it).
- `11.4 s` for the first 200 KiB after a restart → `(n=1)`.
- group `2 MiB in 10.2 s and 25.2 s` → `, n=2`, matching the 8 MiB line right above it.

Re-verified before asking for the merge:

- **The new label cannot lie in the direction it used to.** `falling back to HTTP send` has exactly one emission site (`chat_service.dart:641-644`), reached only when `_trySendChunkedFile` returns false; a successful chunked send returns true before it. `newest_match(..., after=begin[0])` anchors the search after the begin line, so a retry that re-chunks and succeeds is not mislabelled by the previous attempt's fallback.
- **The fallback leg really is HTTP,** so `chunked->http` is truthful: every `begin transfer=` implies the size passed `isChunkCandidate`, and `shouldAvoidWsMonolithicSend` is true at that size, forcing the monolithic retry onto HTTP.
- **Ratios recomputed from the table as written:** `23.0/9.4 = 2.45` (2.4x), `112.0/53.5 = 2.09` (best case 2.1x), `median(53.5, 122.4) = 88`, `112.0/88 = 1.27` (~1.3x). `median(18.2, 20.9, 39.1, 51.1) = 30.0` and `18.2..51.1 / 0.75 = 24..68x` also check out.
- **No stale claim left in the file:** the old "1.7-2.4x slower than one POST" sentence is gone, and the only surviving `path=chunked` mentions are the two that explain the lie and its fix.
- No `lib/` change, no new dependency; the repo has no test convention for the python lab tools (`grep -rn "txlab" test/ tool/` matches only the tool itself), so none was added.

CI green on `638bbef`, CodeRabbit re-ran and reports "Review completed", no open threads.

PR #173, opened 2026-09-07
feat(transfer): same-account multiplatform transfer via backup v3

MrBuondi:
One-active-at-a-time transfer keeping the same onion/ID across Linux/Windows/macOS/Android/iOS.

- Backup v3: hsKeys section inside the same envelope (same password); v2 accepted with new-onion warning.
- HS key export/import: files on desktop, native prysm_tor channel on mobile.
- Restore: HS install + crypto re-init (drop sessions, refresh OTK pool); WAL checkpoint on export, sidecars in manifest.
- Automatic source wipe (HS keys included); Transfer Account UI (en+it).
- Proven live: same-onion on Linux, Android export/wipe, cross-device and cross-platform (Linux to Android).

Deliberately excluded: simultaneous use, group epoch rotation (natural triggers only), Android FilePicker via UI.

## Summary by CodeRabbit

* **New Features**
  * Added account transfer with encrypted backups, hidden-service key preservation, confirmation, and secure data wiping.
  * Added hidden-service key import, export, and clearing across supported platforms.
  * Added detailed restore results and refreshed cryptographic state after transfers.

* **Backup & Restore**
  * Upgraded backups to version 3 with legacy compatibility and database sidecar handling.
  * Added validation and reporting for missing or invalid hidden-service keys.
  * Backup passwords now require at least 8 characters.

* **Bug Fixes**
  * Improved transfer failure reporting, interrupted key installation recovery, and restore reliability.

* **Localization**
  * Added English and Italian transfer and restore messages.


xmreur:
1. Make HS key writes atomic on Android (like iOS): write to .tmp files in hiddenServiceDir, then rename to the final filenames. This prevents torn key files if the process dies mid‑write.
2. Add strict validation before decoding: ensure hostname, hs_ed25519_secret_key, and hs_ed25519_public_key are non‑empty, valid base64, and decode to non‑empty data. Reject early to avoid partial writes.
3. On Windows, _lockDown currently does nothing. Please restrict HS directory/file ACLs to the current user (or at least document the permission gap), since Tor expects key material to be private.
4. Add a warning log when the WAL checkpoint fails. Silent ignore increases the risk of exporting a torn base file.
5. Ensure the native setHsKeys path asserts/enforces that Tor is stopped before writing keys (the comment says so; please enforce it in code).
6. When !result.hasHsKeys, show a dedicated toast explaining the backup is legacy and the onion/ID will change, to set user expectations.
7. Rename hsKeysInstalled to hsKeysInstalledOnDesktop (or add a platform field) since it’s always false on mobile by design.
8. Add tests that pass invalid base64 / missing fields to HsTransferKeys.installToDirectory (and native paths) to confirm rejection without partial writes.

MrBuondi:
Thanks for the review — addressed in ba61e04 (`flutter analyze` clean, `flutter test test/services/backup_service_test.dart` 7/7 pass; native code needs CI to confirm since I can't compile Kotlin/Swift here). Point by point:

1. **Atomic Android writes — done.** `setHsKeys` now stages to `.tmp` files and renames over the finals, matching iOS. Applied the same pattern to `HsTransferKeys.installToDirectory` on desktop, which had the identical torn-write risk.
2. **Strict validation — done.** All three paths (Android, iOS, Dart) now reject empty base64 strings before decoding; decode-then-emptiness checks were already before any write, so bad input never reaches disk.
3. **Windows ACLs — done (best-effort).** `_lockDown` now attempts `icacls /inheritance:r /grant:r <user>:(OI)(CI)F` on the HS dir and both key files, failures ignored. Documented in the method comment that Dart has no ACL API, so a locked-down host account remains the real guarantee.
4. **WAL checkpoint warning — done.** The silent `catch` now logs via `Logging.warning(..., 'BackupService')`.
5. **Enforcing Tor-stopped in native `setHsKeys` — not applied.** Restore installs keys while Tor is running by design and they take effect on the restart the UI already requires (`_performRestore` never stops Tor before installing). A native hard gate would break that flow, turning it into a stop-install-restart dance with more failure modes. I documented the actual contract on `setHsKeysForTransfer` instead. If you want stop-before-install, that's a restore-flow change — happy to do it as a follow-up.
6. **Legacy-backup toast — already present, no change.** `!result.hasHsKeys` already shows `restoreLegacyNoAddressKeys` (en+it: new ID, share it with contacts, restart), distinct from `restoreAddressKeysFailed`.
7. **Rename — done.** `hsKeysInstalled` → `hsKeysInstalledOnDesktop` via LSP rename (service, settings screen, tests).
8. **Rejection tests — done (Dart).** New test covers invalid base64, missing fields, empty strings, and empty-after-decode; asserts rejection, originals untouched, and no `.tmp` leftovers. No unit harness exists for the Kotlin/Swift paths — they share the validate-before-write ordering, covered by review.

MrBuondi:
@xmreur

xmreur:
> @xmreur

you gotta fix coderabbitai errors first

MrBuondi:
@coderabbitai all 4 round-2 findings are addressed in ba64114. cc @xmreur

1. Transfer reports success after shutdown/wipe failure (settings_screen 731-737) — fixed. _performTransfer now tracks stopTor and wipeForTransfer separately and shows the new transferDeactivateFailed string unless both are confirmed. wipeForTransfer returns Future<bool>: false when the native clearHsKeys result is false/unconfirmed (null manager on mobile) or the desktop delete fails.

2. DB handles open during restore (settings_screen 608-611) — fixed. _performRestore now awaits closeForWipe() on DBHelper, MessagesDatabase and PendingMessageDbHelper before restoreBackupDetailed; they reopen lazily, so reinitCryptoForHandoff below gets the restored DB.

3. HS keys installed while Tor runs (settings_screen 608-651) — fixed. _performRestore awaits stopTor() before setHsKeysForTransfer; any failure takes the existing restoreAddressKeysFailed path.

4. HS mutations not serialized (tor_service 172) — fixed. Added one process-wide HsTransferKeys.opMutex: installToDirectory and the new deleteDirectory both run under it, and TorManager set/clearHsKeysForTransfer additionally run under _controlWriteMutex (so HS ops also serialize against start/stop/restart). BackupService (inline desktop install) and PanicWipeService (desktop delete) go through the same helpers, so no direct file mutation bypasses the lock.

Verification: flutter analyze clean on all touched files; backup_service + handoff tests 9/9 (incl. a new deleteDirectory regression test); live-lab round trip on desktop — transfer backup created, source wiped to a fresh onion, restore brought back the original onion, no failure toasts.

xmreur:
@MrBuondi Great work on the transfer feature. Before merging, please address these blocking issues:

## Blocking changes required

1. **Atomic source deactivation (Tor stop + HS-key deletion)**
   - Merge `stopTor()` and `clearHsKeysForTransfer()` into a single `TorManager`-owned operation (e.g. `deactivateForTransfer()`).
   - Acquire the lifecycle/control mutex, suppress automatic restarts, stop and verify Tor is dead, delete HS keys under the same lock, verify their absence, then release.
   - `PanicWipeService.wipeForTransfer()` should call this one operation instead of deleting desktop key files directly.

2. **Atomic HS-key installation (all-or-nothing)**
   - Android/iOS/desktop currently stage files and rename secret → public → hostname in sequence; a failure after the first rename leaves a mixed key set.
   - Use directory-level staging with an atomic swap where supported, or rollback/remove all three destination files on any commit failure.
   - Add a postcondition check that exactly the full triplet exists before returning success.

3. **Android `clearHsKeys()` result handling**
   - `deleteRecursively()` and `mkdirs()` return Booleans that are ignored; deletion failure can leave keys behind.
   - Return `false` when deletion fails and verify the secret-key file does not exist before reporting success.

## Additional recommendations

- Strengthen backup password policy for all backups (this archive includes the onion private key).
- Use a dedicated success message for transfer (e.g. "Transfer completed. This device has been deactivated; restart or close the app.").
- Resolve the failing `github-advanced-security` check.

Once these are fixed (and ideally covered by tests simulating restart races and partial-write failures), I'm happy to approve.

MrBuondi:
@xmreur all three blocking issues are addressed in ce233a4:

**1. Atomic source deactivation.** New TorManager.deactivateForTransfer() owns the whole sequence under _controlWriteMutex: it sets a fail-closed flag first (checked at the top of startTor/restartTor, so supervisor auto-restarts, controller restarts and the UI button all refuse until app restart), then stop, verify dead (desktop: process handle null + control-port connect-refused probe; mobile: the native stopTor already awaits port-close and throws on failure), HS delete, and triplet-absence verify. PanicWipeService.wipeForTransfer() now calls this one op (direct desktop delete kept only as a no-manager fallback). Live proof: after wipe the HS dir is gone and no Tor process comes back — previously one relaunched and minted a fresh onion.

**2. Atomic HS install.** All three installers now distinguish staging from commit: a staging failure leaves pre-existing live keys untouched (an adversarial review caught my first version deleting them — fixed); only a post-promote failure rolls back to no keys, and a triplet postcondition (all three files exist, non-empty) gates success. So the outcome is always complete-old, complete-new, or absent — never mixed.

**3. Android clearHsKeys.** deleteRecursively/mkdirs results are checked and the full triplet (not just the secret file) must be absent before returning true. Same triplet check added to iOS clearHsKeys.

**Recommendations:** backup password minimum raised 4→8 chars for transfer and regular backups (passwordMustBeAtLeast8Characters, en+it); new dedicated success message transferCompletedDeactivated ("Transfer completed. This device has been deactivated; restart or close the app."); the advanced-security check is green on the latest run — the failure you saw did not reproduce.

**Verification:** flutter analyze clean; 12/12 backup+handoff+deactivation+sidecar tests pass plus 30/30 Tor suites; new transfer_deactivation_test covers deactivate+verify, restart suppression, wipe propagation, and an install/delete race asserting whole-states-only. Live-lab round trip: 8-char policy toast, backup, wipe with the new message, source stays down, restore brings back the transferred onion.

**Caps, stated openly:** the Kotlin/Swift changes are carefully re-read but not compiled here (no Android/iOS toolchain in this environment) — worth a CI/local build check. iOS restartTor native path is byte-identical to base (verified via diff).

MrBuondi:
@coderabbitai addressed in f282b54. cc @xmreur

**1. Repeat the transfer guard inside both protected callbacks (tor_service.dart) — fixed.** Confirmed the race: the outer check ran before `_controlWriteMutex`, so a start queued behind a pending deactivation could still run after key deletion. `deactivateForTransfer()` now sets `_deactivatedForTransfer` synchronously *before* queueing, and both `startTor` and `restartTor` re-check it as the first statement inside their protected callbacks.

**2. Validate triplets before native dispatch / reject malformed non-empty key files — fixed with one shared validator.** `HsTransferKeys.isValidTriplet` checks the zero-padded `== ed25519v1-secret/public: type0 ==` headers and the 96/64-byte lengths, then recomputes the v3 onion from the public key (SHA3-256 checksum + base32 via pointycastle, already a dependency) and requires the hostname to match. Used on the desktop install path and by `isValidEncodedTriplet` before native `setHsKeys` dispatch, so a base64-clean but malformed or mismatched triplet now fails the restore instead of silently installing an unusable onion.

**3. Crash-recoverable install (Dart + Android + iOS) — fixed via marker + startup repair, not directory swap.** Accepted the finding: per-file rollback cannot run if the process dies between promotes. Not using an "atomic directory swap" because POSIX `rename()` fails with ENOTEMPTY on a non-empty destination directory, so that approach still needs two renames and only moves the window. Instead a `.hs_install_pending` marker is written before the first promote and removed only after the triplet postcondition passes; `startTor` on all three platforms purges the possibly-mixed set when the marker survived, falling back to a fresh onion (recoverable — the user restores the backup again).

**Not doing, with reasons:** "replace placeholder fixtures with triplets generated by each supported Tor binary" — the new `test/support/hs_fixtures.dart` derives the hostname from the public key with the real algorithm, which gives format-valid triplets without per-platform Tor binaries; and an Android instrumentation test that kills the process after each promotion — the marker makes the repair deterministically testable without simulating process death.

**Verification:** analyze clean; 13/13 backup + deactivation tests, 32/32 across Tor/panic/handoff suites. Live desktop lab, real Tor-generated keys: transfer → "Transfer completed" → HS dir gone and no Tor process returns; restore → the *real* onion `vjfpsi43…` reinstalled and validated (so the new validator accepts genuine Tor output end-to-end); then marker planted by hand + app restart → triplet purged, fresh onion `uubltet7…` minted, exactly the intended repair. Cap unchanged: Kotlin/Swift re-read but not compiled here.

MrBuondi:
@coderabbitai addressed in ffb6ce9. cc @xmreur

**1. Fail-closed repair (Dart `repairInterruptedInstall`, Kotlin/Swift `repairInterruptedHsInstall`) — fixed, all three platforms, same protocol.** Purge attempts every live and staged key file, verifies every one is absent, and only then removes `.hs_install_pending`. If a file survives, the marker is kept and the start is refused: desktop `startTor` **and** `restartTor` throw `StateError`, Android `startTor` throws `IllegalStateException`, iOS `repairInterruptedHsInstall()` now `throws` and `startTor()` propagates it. Install rollback (`_installUnlocked`, Kotlin `rollbackHsInstall`, Swift `rollbackHsInstall`) uses the same purge-then-marker order.

**2. Marker removal on the success path (Kotlin L262, Swift L300) — fixed.** A failed marker delete after a valid promote now returns `false` instead of a success the next start would undo. Rationale for keeping the marker rather than trying to "preserve the triplet": with the marker present the next start purges deterministically, and the user has already been told the restore failed — a consistent, recoverable state.

**3. Lockdown before staging (Dart L276) — fixed.** The HS dir is `chmod 700` **before** any secret byte is staged (a 0700 dir shields the files whatever the umask gives them), key files `chmod 600` after promote, and on POSIX both the chmod exit code and the resulting `FileStat.mode` are checked — failure is an install failure. Marker removal is now the last step, after lockdown. Windows `icacls` stays best-effort: Dart has no ACL API to verify the result, and `%APPDATA%` is already user-private by inheritance, so a hard failure there would only turn a non-regression into a broken restore.

**Not doing: secret↔public correspondence in `isValidTriplet`.** Tor is the authoritative validator here — `loadkey.c` derives the public key from the secret (`ed25519_public_key_generate`) and on mismatch logs *"does not match … If you are trying to restore from backup, make sure you didn't mix up the key files"* and `goto err`, so a mismatched pair can't run as an "unusable identity"; it fails loudly at start. The triplet is Tor's own directory read verbatim inside the AEAD-authenticated backup, so a mismatch is unreachable without breaking the envelope. And nothing in the dependency tree exposes expanded-scalar Ed25519 basepoint multiplication (`cryptography` only takes seeds; its `_pointMul` is private; pointycastle has no Ed25519), so this would be ~50 lines of hand-rolled curve arithmetic guarding a path Tor already guards.

**Regression test (as requested):** `transfer_deactivation_test` plants the marker, makes the HS dir read-only so unlink fails, and asserts repair returns `false`, marker + keys survive, and `startTor`/`restartTor` both throw. Fails on f282b54 (`Expected: false, Actual: <true>`), passes on ffb6ce9. Skips itself when running as root.

**Verification:** analyze clean; 6/6 deactivation, 15/15 backup+handoff+deactivation, 126/126 across the Tor suites. Live Linux lab, real Tor keys, end to end: transfer → source deactivated (HS dir gone, no Tor process); restore → the original onion reinstalled under `0700`/`0600`, Tor restarts on it; then marker planted + `chmod 500` → app refuses to start Tor (`Tor initialization failed: Bad state: Interrupted hidden-service install could not be purged`), marker and triplet intact; `chmod 700` + in-app Retry → purge, fresh onion. Cap unchanged: Kotlin/Swift re-read (plus an adversarial review pass that caught the desktop `restartTor` bypass) but not compiled here.

MrBuondi:
@coderabbitai iOS temp-file gap addressed in the follow-up commit. cc @xmreur

`setHsKeys(_:)` now stages to deterministic `hs_ed25519_secret_key.tmp` / `hs_ed25519_public_key.tmp` / `hostname.tmp`, writes the marker, promotes with POSIX `rename(2)` (atomic replace over an existing final file — `FileManager.moveItem` refuses that), and `purgeHsTriplet()` removes **and verifies absent** both the final and `.tmp` names before `.hs_install_pending` is removed. Same protocol as Android and desktop now. One extra safeguard while there: a staging failure (before any rename) also removes a possibly-partial marker, so the next start cannot purge the untouched pre-existing keys.

Cap unchanged: Swift re-read carefully, not compiled here.

xmreur (approved):

PR #175, opened 2026-09-13
Relay v1: store-and-forward for offline peers, plus cheap creation and one-step pairing

MrBuondi:
Supersedes #174, which was merged and then rolled back: `main` is at `2d8128a` and carries none of this work, so this PR contains the whole Relay effort — the original v1 (40 commits, unchanged) plus nine new commits that make a relay cheap to create, cheap to run and one step to pair.

## What a Relay is

A Relay holds sealed messages for an offline owner until that owner asks for them. It is not a central server: anyone can run one, nobody needs one, and two peers online together never touch one. It cannot read what it stores, it never sends on your behalf, and it is reachable only as a Tor onion service. The normative wire contract is `.scratch/relay/proto/relay-protocol-v1.md` (`prysm-relay/1`); it is **unchanged by this PR**.

## What is new since #174

**Creating a relay.** Provisioning used to compile the binary on the operator's machine (Dart SDK required), pull `ubuntu:24.04` and `apt install tor` inside a bare container, driven by 562 lines of bash. Now a prebuilt multi-arch image does the work and `create_relay.sh` is a 199-line wrapper around `docker run`; for a dedicated host without the app there is a native installer plus a systemd unit, which is the primary path. Measured: image 136 MB, `docker run` → `listening on` in 5 s, 9.1 MiB RSS for the whole container.

**Two defects that cost data or availability.**
- Persistence was opt-in, so `docker rm` took `identity.json` and the hidden-service key with it and every signed Contract died. Volumes are now always mounted; `--ephemeral` keeps the old behaviour behind an explicit warning.
- `docker run -d` carried no restart policy, so a reboot, a daemon restart or a crash left the relay down **in silence** — a dead relay is indistinguishable from a relay with nothing to deliver. Now `--restart unless-stopped`, proven by killing `serve` inside the container and watching Docker bring it back with the same onion.

**Pairing in one step.** The owner used to copy an onion, a fingerprint and a token by hand, and compare hex across two screens. `prysm_relay pair-link` now prints a `prysm-relay://pair?onion=…&fpr=…&token=…` link and the same link as a scannable QR block; the app pastes, auto-detects or scans it, fills the form, starts the fetch and — because the link carries the fingerprint — **checks the relay's identity itself**, blocking Pairing on a mismatch. The link is a provisioning artifact, not wire: no signature, no timestamp, nothing reaches the relay, and it is as secret as the token inside it.

**The cold-onion trap.** The first request to a fresh onion takes longer than the app's 30 s budget (measured 39.8 s), which used to end in a mute error on a perfectly good relay. One automatic second attempt with a 60 s budget, with an explicit "Still connecting…" message; a relay that answers with an error is never retried.

**Low-power hosts.** The 60 s sweeper is now pinned by a test to write nothing when there is nothing to expire (SD-card wear on a Raspberry), the Tor runbook gained a low-power profile, and the README states the hardware floor: armv7+/arm64/x86_64, 256 MB free RAM (armv6 — Pi Zero W, Pi 1 — excluded, Dart needs armv7).

**Release and CI.** `relay.yml` is path-filtered and Dart-only, and runs the race tests three times on Linux as the rule requires; `relay-release.yml` publishes, on a `relay-v*` tag, x64/arm64/armv7 tarballs with checksums and keyless provenance plus a multi-arch GHCR image with an SBOM, and rebuilds `:latest` weekly so Tor security updates land without manual work. `ci.yml` now **fails** if `prysm_relay_server` is ever referenced from `lib/`, `test/` or `pubspec.yaml`, a rule that previously existed only as a comment.

**Podman.** Verified on 6.1.1 rootless: the script needs no change, but `alias docker=podman` cannot work — an alias is not inherited by a child process, so the script keeps calling the real docker binary and the operator provisions on Docker while believing otherwise. Documented with the PATH shim that does work, and with the three real differences (fully qualified image names, `podman-restart.service` for reboots, rootless volume paths).

## Verification

- `flutter analyze` clean; **1280** app tests green (1 skipped: the gated real-Tor harness).
- `prysm_relay_protocol` **30/30**; `prysm_relay_server` **53/53**, run three times per the race rule.
- Container smoke on a throwaway name and port: four pairing lines plus QR, `listening on`, one token spent per boot, teardown clean. The QR block was rasterised and decoded with `zbarimg`: byte-identical to the `link:` line.
- **Live app test** (the project rule for any client change): typing the link filled both fields and started the fetch; the first attempt timed out and showed the retry text; the second rendered the preview with `Fingerprint matches the pairing link.`; Pair completed in 27 s; a link with one altered hex char showed the mismatch with Pair disabled; a malformed link showed the invalid-link banner; no errors in the app log.
- Not exercised: QR *scanning* is gated to Android/iOS, so the Linux lab cannot drive it (a widget test asserts the gating); the release workflows run for the first time on the first `relay-v*` tag.

## Notes for the reviewer

- The nine new commits are grouped by theme and each message carries the defect, the fix and the measurement. The first 40 are #174's, unchanged.
- `docs/HANDOFF-relay-2026-09-12.md` is untracked by the last commit: session handoffs are notes between working sessions, not project documentation. The file stays on disk, out of git, next to the others.
- After merge, publishing the image needs one manual step: a GHCR package starts **private** even on a public repository, and making it public is irreversible.


MrBuondi:
Split for review: CodeRabbit skipped this PR (125 files, limit 100) because `main` was rolled back after #174 merged, so the 40 already-reviewed commits are back in the diff. The nine new commits are now split into a stack sized for review — #176 (server: pairing link, image, native install, Podman, low-power), #177 (app), #178 (ci/release), #179 (docs, tracker, handoffs out of git) — each based on the previous one, on top of `relay-v1-reviewed` (= the tip reviewed in #174). Those four are review-only; this PR stays the one that carries the whole effort to `main`.

